Issue #20, opened 2019-06-13
doc build and dependencies


hjoliver:
#3 moved all docs here from the cylc-flow repository, simplistically taking the relevant bits of `setup.py` and `cylc-check-software` with it.

We need to decide how best to build the docs here, and how best to depend (where necessary) on other project/repo versions, particularly cylc-flow.  (e.g. should we use `setup.py` here at all?)

See also #17

kinow:
Related to this one, we have an issue in `setup.py` that prevents the project from building in a new virtual environment. We have `sphinx` in `install_requires`. And we also do `import sphinx` in `setup.py`. The problem is that the import in `setup.py` is evaluated first, resulting in an `ImportError`.

Using `setup_requires` could solve it, if that was evaluated by pip. We could alternatively hack `setup.py` and install sphinx if not present. But I think that depends first on the decision pending in this issue.

oliver-sanders:
closed by #26

PR #25, opened 2019-06-27
Remove (now cross-repo) references to minor files

sadielbartholomew:
When building the docs (as per the instructions in the ``README.md`` here) I get a few Sphinx warnings. I'm using ``Sphinx v2.0.1``, but the nature of the warnings mean they should appear for any version; they both arise because there are references for raw inclusion (via the ``literalinclude`` directive) of some files that are still in, & belong, in the ``cylc-flow`` directory, so are no longer in this repository (see also #20).

Fortunately, these are all minor &mostly unhelpful files, so are easy to deal with; no cross-repo referencing or duplication is required:

* ``INSTALL.md`` was very out of date (I will open an Issue to register this) & there is an Installation section for this anyway :arrow_right: remove the section which just prints it;
* ``README.md`` is viewable more accessibly & agreeably in markdown via the GitHub codebase & currently pointing to the wrong file (the one in ``cylc-flow`` is the appropriate one here) :arrow_right: ditto;
* the example suite definition, ``flow.rc.eg``, outlines a small amount of important information regarding the ``flow.rc`` (view that file [here](https://github.com/cylc/cylc-flow/blob/master/cylc/flow/etc/flow.rc.eg)) :arrow_right: just place that content (formatted to RST) into the docs proper, in the same location it was previously included.

With the above amendments, the warnings naturally are resolved & no longer appear. We then achieve a clean docs build.

### DIFF
--- a/doc/src/global-flow-conf.rst
+++ b/doc/src/global-flow-conf.rst
@@ -6,10 +6,31 @@ Global (Site, User) Configuration Files
 Cylc site and user global configuration files contain settings that affect all
 suites. Some of these, such as the range of network ports used by cylc,
 should be set in the site ``flow.rc`` config file. Legal items,
-values, and system defaults are documented in (:ref:`SiteRCReference`).
+values, and system defaults are documented in :ref:`SiteRCReference`.
 
 Others, such as the preferred text editor for suite configurations,
-can be overridden by users in ``~/.cylc/flow/<CYLC_VERSION>/flow.rc``
+can be overridden by users in ``~/.cylc/flow/<CYLC_VERSION>/flow.rc``.
 
-.. literalinclude:: ../../cylc/flow/etc/flow.rc.eg
-   :language: none
+
+How to create a site or user flow.rc config file
+------------------------------------------------
+
+The ``cylc get-global-config`` command prints global config defaults,
+overridden by site global settings (if any), overridden by user global
+settings (if any).
+
+To generate a new global config file:
+
+1. Run ``$ cylc get-global-config > flow.rc``
+2. Edit any settings that you need to modify.
+3. Delete or comment out any seeting you do not need (to avoid inadvertently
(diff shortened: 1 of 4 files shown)


doc/src/global-flow-conf.rst
```diff
@@ -6,10 +6,31 @@ Global (Site, User) Configuration Files
...
+settings (if any).
+
+To generate a new global config file:
+
+1. Run ``$ cylc get-global-config > flow.rc``
+2. Edit any settings that you need to modify.
+3. Delete or comment out any seeting you do not need (to avoid inadvertently
+   overriding defaults or site settings that may change in the future).
+
+For all legal items, see the :ref:`SiteRCReference`.
```
kinow: s/legal items/valid items?

sadielbartholomew: Sorry @kinow, I am not sure what this comment means. Can you please explain?

kinow: Was just wondering if valid would sound better than legal? Although normally I use illegal item.

kinow: Resolving as I think legal is fine

sadielbartholomew: Good thought. However I did not actually write this content, I copied it from the file that was previously included in raw format here (bar for making some formatting changes to sit right in RST format), i.e. the ``flow.rc.eg`` file, i.e. [this one](https://github.com/cylc/cylc-flow/blob/master/cylc/flow/etc/flow.rc.eg), so I wouldn't consider it in the scope of this PR to revise the text there.

I really just want to make sure the build warnings are addressed without losing any content that was useful & there before.

hjoliver: (I'm the original author of the aforementioned line; valid might be better, but I doubt anyone will misunderstand legal).

PR #31, opened 2019-07-09
pycodestyle

oliver-sanders:
~**Built atop #26, merge after.**~

Run the pycodestyle linter over all Python code.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,9 +38,10 @@ env:
 
 install:
   - pip install git+https://github.com/cylc/cylc-flow/
-  - pip install -e .
+  - pip install -e .'[all]'
 
 script:
+  - pycodestyle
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -27,26 +27,22 @@
...
-addons:
-  apt:
-    update: true  # instruct travis-ci to always run apt-get before each build
 
 language: python
-
-stages:
-- documentation
+python:
+  - "3.7"
```
kinow: Funny that other PRs failed due to missing sphinx module (https://github.com/cylc/cylc-doc/issues/32) but not this one.

Maybe py 3.7 in travis includes the sphinx module?

oliver-sanders: Probably because I've added `pip install -e .` to the `.travis.yml`?

kinow: Ah, that makes sense!

PR #42, opened 2019-07-29
Add ISODATETIMEREF and ISODATETIMECALENDAR to job script.

matthewrmshin:
Companion of cylc/cylc-flow#3249.

### DIFF
--- a/src/suite-config.rst
+++ b/src/suite-config.rst
@@ -2281,6 +2281,9 @@ The task job script may export the following environment variables:
    CYLC_VERSION                    # Version of cylc installation used
 
    CYLC_CYCLING_MODE               # Cycling mode, e.g. gregorian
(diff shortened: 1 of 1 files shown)


src/suite-config.rst
```diff
@@ -2281,6 +2281,9 @@ The task job script may export the following environment variables:
    CYLC_VERSION                    # Version of cylc installation used
 
    CYLC_CYCLING_MODE               # Cycling mode, e.g. gregorian
+   ISODATETIMECALENDAR             # Calendar for the `isodatetime`,
+                                   # defined with the value of CYLC_CYCLING_MODE
+                                   # in a date-time cycling mode
```
oliver-sanders: Perhaps note that this is for the `isodatetime` command.

matthewrmshin: Done.

PR #38, opened 2019-07-23
Cylc Tutorial

oliver-sanders:
close #18 
address #11 
if travis passes - fix #32 

Transfer the Cylc Tutorial and all of its dependent scripts from the Rose Documentation.

* Copy content from Rose
  * Glossary
  * Cylc Tutorial
* Graphics
  * Remove unused graphics
  * Use SVG graphics directly
    * No need to convert to bitmap - Sphinx can do this for us
    * Scaled graphics didn't seem to be used
  * Remove conversion script
  * Convert Cylc logo to SVG (see cylc/cylc-ui#129)
* Sphinx
  * Change documentation theme (required by Sphinx extensions)
  * Fix issue with versioning caused by change in `cylc --version` output
  * Slides builder
* Random
  * Run eslint over javascript files

Tried to avoid styling the docs but there is a little in: 2637357, bb7e104. Can revert if controversial.

Otherwise a fairly straight-forward lift and shift. Most code has already been reviewed in Rose, look at individual commits to see changes.

PR #43, opened 2019-07-29
update cylc_lang extension

oliver-sanders:
Address syntax highlighting part of https://github.com/cylc/cylc-doc/pull/38#pullrequestreview-267514649.

Turns out that the `cylc_lang` extension in cylc-doc had got out of date with the one in rose.

The cylc-doc repository should be seen as the canonical location for Cylc Sphinx extensions. In the future we may remove it from Rose and find some way to import it from cylc-doc.


oliver-sanders:
For reference this diff is the last commit of https://github.com/metomi/rose/pull/2274

PR #50, opened 2019-08-02
Document xtrigger template escape.

hjoliver:
What it says.  Ref: https://github.com/cylc/cylc-flow/pull/3258


kinow:
Travis failure not related to this change.

kinow:
Accidentally double-approved (didn't know it was possible without any reply/changes heh).

One review will do IMO

PR #52, opened 2019-08-07
restrict suite name character set

oliver-sanders:
Companion to https://github.com/cylc/cylc-flow/pull/3274

### DIFF
--- a/src/suite-name-reg.rst
+++ b/src/suite-name-reg.rst
@@ -3,9 +3,15 @@
 Suite Name Registration
 =======================
 
-Cylc commands target suites via their names, which are relative path names
-under the suite run directory (``~/cylc-run/`` by default). Suites can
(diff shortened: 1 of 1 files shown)


oliver-sanders:
Tests will pass once the Cylc Flow PR is merged

src/suite-name-reg.rst
```diff
@@ -3,9 +3,15 @@
 Suite Name Registration
 =======================
 
-Cylc commands target suites via their names, which are relative path names
-under the suite run directory (``~/cylc-run/`` by default). Suites can
-be grouped together under sub-directories. E.g.:
+Cylc commands identifies suites via their names, which are relative path names
```
hjoliver:
```suggestion
Cylc commands identify suites via their names, which are relative path names
```

PR #57, opened 2019-08-21
Remove numbering for listings w/ captions & figures

### LINKED ISSUE #44: drop numbered figures
Remove the figure numbering inherited from the LaTeX docs.

See "code block labels" comment in https://github.com/cylc/cylc-doc/pull/38#pullrequestreview-267514649 and the related #15

sadielbartholomew:
Close #44. There are two components to this, where numbering has been removed for:

1. (as per the Issue to close) all figures as included via the ``figure:`` directive;
2. a small number of cases of code blocks numbered via 'Listing N' for some integer N, resulting from any use of the ``:caption:`` directive, which adds 'Listing N' before the stated text (e.g. ``Listing 1 input`` for ``:caption: input`` in a code block).

I have also (3) moved for ``suite.rc`` code with accompanying ``cylc graph`` output, the graph to the end rather than the start, as it seems more helpful & natural to show the code before showing the graph produced by it. It also means the caption is at the top, rather than in the middle, which had the potential to be slightly confusing.

Further context is provided below for each case, if helpful to understand the code changes. For reviewing, please build the docs & view the pages with figures (see ``git grep "figure\:`` & ... ``:caption:``) to check they are incorporated in a seamless way into the text.

#### 1) Figures

Removing numbering from the figures is trivial (just remove ``numfig = True`` etc. in the ``conf.py`` & change all ``:numref:`` to a standard ``:ref:`` directive), but it is not trivial to do so while ensuring the figures still are referenced properly & sit in the most appropriate place relative to the content, e.g. below or close to the text they relate to.

With the 'trivial' change above, Sphinx makes all references via ``:ref:`` to figures display as the text linking to the figure location, instead of numbers, the full caption, e.g. here for *"'A single cycle point ...' shows the optimal job ..."*:

caption_example_scr

Naturally this ruins the flow of the text, so we can't leave figures referenced in that way. The only natural solution, which is the ideal at least IMO anyway, is to:

* integrate the figures into the dialogue, such that relevant text references the figure & introduces it with a colon, or similar.
* for any further (non-linear) references elsewhere to the same figure, use a reference where the linked text is some custom test that describes the figure in context.

This involved moving some figures up or down relative to the text, & either removing to become explicit textual references, or amending as appropriate, the ``:ref:`` references.

#### 2) Code blocks with a ``:caption:``

From investigation & then a ``grep ``to pick out all ``:caption:`` instances, there are only a small number of such cases.

Since we only use a small number of code block captions, & they seem unnecessary not just for the current instances as changed here but in general, I am just moving the necessary text out of captions & into the preceding text.

Now the cases look like so (compare with the screenshots in https://github.com/cylc/cylc-doc/pull/38#pullrequestreview-267514649):

A1_scr

A1_scr2

#### 3) Some movements

Before such cases had this format:

caption_2_scr

The changes here make the caption come first (in italics), then the ``suite.rc`` code, then the resultant ``cylc graph`` output.

### DIFF
--- a/src/suite-config.rst
+++ b/src/suite-config.rst
@@ -265,25 +265,18 @@ vary depending on the particular cycle point:
            # hours
            T06,T18 = "C => X"
 
-:numref:`fig-dep-eg-1` shows the complete suite.rc listing alongside
-the suite graph. This is a complete, valid, runnable suite (it will
-use default task runtime properties such as ``script``).
+Here is the complete suite.rc listing with the corresponding suite graph;
+this is a complete, valid, runnable suite (it will use default task runtime
+properties such as ``script``):
 
 .. Need to use a 'container' directive to get centered image with
    left-aligned caption (as required for code block text).
 
 .. container:: twocol
 
-   .. container:: image
-
-      .. _fig-dep-eg-1:
-
-      .. figure:: img/dep-eg-1.png
-         :align: center
-
    .. container:: caption
 
-      Example Suite
+      *Example Suite*
 
       .. code-block:: cylc
 
@@ -307,6 +300,13 @@ use default task runtime properties such as ``script``).
              final cycle point = 20200401T06
              [[node attributes]]
                  X = "color=red"
+
+   .. container:: image
+
+      .. _fig-dep-eg-1:
+
+      .. figure:: img/dep-eg-1.png
+         :align: center
 
 
 Graph String Syntax
@@ -466,7 +466,7 @@ to generate the final graph.
 One-off (Non-Cycling)
 """""""""""""""""""""
 
-:numref:`fig-test1` shows a small suite of one-off non-cycling
+The following is a small suite of one-off non-cycling
 tasks; these all share a single cycle point (``1``) and don't spawn
 successors (once they're all finished the suite just exits). The integer
 ``1`` attached to each graph node is just an arbitrary label here.
@@ -476,16 +476,9 @@ successors (once they're all finished the suite just exits). The integer
 
 .. container:: twocol
 
-   .. container:: image
-
-      .. _fig-test1:
-
-      .. figure:: img/test1.png
-         :align: center
-
    .. container:: caption
 
-      One-off (Non-Cycling) Tasks.
+      *One-off (Non-Cycling) Tasks*
 
       .. code-block:: cylc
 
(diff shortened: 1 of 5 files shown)


sadielbartholomew:
Sorry given you have already reviewed, @kinow, but I am expanding the scope of this PR from the original PR changes as the nature of the Issue at hand, #44, was not clear, & it turns out I have not yet addressed it, when my main aim is to close it.

kinow:
No problemo @sadielbartholomew , haven't built the cylc-doc project in a while, so will re-review and meanwhile take a look if there are any warnings or anything strange in the build process :+1: thanks!

src/suite-config.rst
```diff
@@ -265,7 +265,7 @@ vary depending on the particular cycle point:
            # hours
            T06,T18 = "C => X"
 
-:numref:`fig-dep-eg-1` shows the complete suite.rc listing alongside
+:ref:`fig-dep-eg-1` shows the complete suite.rc listing alongside
```
kinow: Failing in Travis?

```
Warning, treated as error:

/home/travis/build/cylc/cylc-doc/src/suite-config.rst:268:undefined label: fig-dep-eg-1 (if the link has no caption the label must precede a section header)

Makefile:31: recipe for target 'html' failed
```

sadielbartholomew:
Oh sorry @kinow this was not ready for re-review yet. I think I pushed up the last commit by mistake.

sadielbartholomew:
Ready for re-review, with the extended scope which I have just amended my opening comment to reflect.

src/tutorial/runtime/introduction.rst
```diff
@@ -81,14 +81,16 @@ We can also call other scripts or executables in this way, e.g:
    Cylc will automatically prepend it to the :envvar:`PATH` environment
    variable when the task runs.
 
+``bin/hello_world``:
+
 .. code-block:: bash
-   :caption: bin/hello_world
```
oliver-sanders: Can leave these captions unchanged.

oliver-sanders:
> I used to be able to open the docs locally

I'm guessing this will be due to `@fontface` stuff in the CSS, take a look at the developer console (enable css logging) you may see a cross-size warning or something of that effect.

src/tutorial/runtime/introduction.rst
sadielbartholomew: Without the change they are getting numbering e.g. 'Listing N: bin/hello_world' written out above. Just to check, you are aware of that & fine with it in this case?

oliver-sanders: "Listing N" was being added to the captions `numfig` config which this PR has removed.

sadielbartholomew: Ah! I did not realise that. Thanks for the pointer. I'll revert the changes to pull out the previously-affected code block captions in a new commit, then, then squash that down.

sadielbartholomew:
I have just updated this: after first review there were two commits, the [first addressing case 2 & the second case 1](https://github.com/cylc/cylc-doc/pull/57#issue-309650312), & all I have done is:

* do a ``git revert`` on the original first commit;
* then a ``rebase`` to squash the 2 + 1 reverted commits down so the one commit now there reflects only the second commit as previously reviewed.

to ultimately drop the first commit, which addressed case 2 by amending the use of directives in relevant places, but it was already inherently achieved by the removal of the enabling of ``numfig`` in the following commit (see https://github.com/cylc/cylc-doc/pull/57#discussion_r318584440).

oliver-sanders:
@kinow would you quickly eyeball this as it is now quite different from the original PR.

kinow:
Looks good, had a look at the forced commit, just captions added. LGTM

PR #65, opened 2019-08-25
Update echo xtrigger in line with current (Python 3) code

sadielbartholomew:
The built-in "custom" xtrigger ``echo`` is directly written out in the docs, & is now out of date, as the [actual xtrigger code](https://github.com/cylc/cylc-flow/blob/283bda1da40ee0d3baa14556000f4d6a4e3f4d43/cylc/flow/xtriggers/echo.py#L28-L29) (see also the [xtriggers directory equivalent](https://github.com/cylc/cylc-xtriggers/blob/a83eb011021cf67f96e79c14f4e003fa7d3f2ab8/cylc/flow/xtriggers/echo.py#L27-L28)) was updated from Python 2 to 3.

Ideally, instead of just updating the written-in code to match (actually, strictly I made it match the xtrigger repo version without the tuple parentheses in the return value, as they are not required), we could use a ``literalinclude:`` directive to add in the code itself, ensuring it will stay up-to-date, but I am not sure if that would be too much hassle to set-up, given:

* it requires connecting up to the ``cylc-flow`` repo to access the ``echo.py`` file;
* there is only one xtrigger example shown, & this ``echo`` example is so trivial it is unlikely to change again.

Are we happy with this "simple content update" approach for now? If not, let me know, & I can amend this accordingly.

### DIFF
--- a/src/external-triggers.rst
+++ b/src/external-triggers.rst
@@ -298,7 +298,7 @@ Toy Examples
 echo
 """"
 
-The trivial built in ``echo`` function takes any number of positional
+The trivial built-in ``echo`` function takes any number of positional
 and keyword arguments (from the suite configuration) and simply prints
 them to stdout, and then returns False (i.e. trigger condition not
 satisfied). Here it is in its entirety.
@@ -306,9 +306,9 @@ satisfied). Here it is in its entirety.
 .. code-block:: python
 
    def echo(*args, **kwargs):
-       print "echo: ARGS:", args
-       print "echo: KWARGS:", kwargs
-       return (False, {})
+       print("echo: ARGS:", args)
+       print("echo: KWARGS:", kwargs)
+       return False, {}
 
 Here's an example echo trigger suite:
 
@@ -332,7 +332,7 @@ your terminal).
 xrandom
 """""""
 
-The built in ``xrandom`` function sleeps for a configurable amount of
+The built-in ``xrandom`` function sleeps for a configurable amount of
 time (useful for testing the effect of a long-running trigger function
 - which should be avoided) and has a configurable random chance of
 success. The function signature is:


kinow:
>it requires connecting up to the cylc-flow repo to access the echo.py file;

All good for me, as we wouldn't need to update doc after updating src.

I think there is a dependency already between cylc-doc and cylc-flow.

But not in setup.py... I am not in front of the computer, but I suspect you need cylc-flow to build cylc-doc? Or I could be wrong... will check when I get to the office in a few minutes

sadielbartholomew:
Thanks @kinow, though please don't go out of your way to check, it is non-urgent. Take your time.

kinow:
No worries @sadielbartholomew , was going to take a break from the Tree component and work on other issues this week anyway :)

I've created a blank virtual environment, and installed via `pip install .`, followed by `make html`. You immediately get an error

```
ModuleNotFoundError: No module named 'cylc'
```

So cylc-doc depends on cylc-flow, but I believe the intention is not to have the cylc-flow in `setup.py`, and instead leave it up to the developer to make sure the version of cylc-flow against which the docs must be generated is correctly installed before running `make` - https://github.com/cylc/cylc-doc/pull/38#discussion_r306636920

I guess you can use the `literalinclude:` then, as long as you remember to install cylc-flow first :+1:

sadielbartholomew:
Ah, thank you very much for your explanations, kinow, that's really helpful.

> instead leave it up to the developer to make sure the version of cylc-flow against which the docs must be generated is correctly installed before running make - #38 (comment)

TBH I'm not entirely happy with that approach, it seems risky & extra work in terms of maintenance to me, but if that was the decision made I will go with it.

I'll squash in another commit to change this PR to update via the following approach:

> I guess you can use the literalinclude: then, as long as you remember to install cylc-flow first

kinow:
>TBH I'm not entirely happy with that approach, it seems risky & extra work in terms of maintenance to me, but if that was the decision made I will go with it.

I also prefer to be explicit about dependencies (for other devs and users). But am not entirely familiar with Sphinx or the whole build of cylc-doc (understand more the `setup.py` version to be honest).

hjoliver:
Thanks @sadielbartolomew :+1: this hard-wired doc change can go in as-is, as it is an improvement.  We can consider literal include and build dependency elsewhere if we need to...

PR #94, opened 2019-10-21
document coreutuils dependency

oliver-sanders:
Whilst this section will be re-written anyway we should make a note of our `coreutils` dependency.

Coreutils is required for basic functionality (e.g. the `timeout` command used within Cylc Flow) and optional functionality (e.g. the functional tests which use non-POSIX stuff like `sed -r`).

Note: We need Coreutils to be installed "as normal", e.g. on BSD it is safer to install commands with a prefix (e.g. `ls` => `gls`) to avoid conflict with the BSD equivalents. Typical workaround is to fudge it in the `.bashrc` file. Will add instructions for this when deployment is sorted.

### DIFF
--- a/src/installation.rst
+++ b/src/installation.rst
@@ -31,8 +31,10 @@ The following packages are necessary for running tests in Cylc:
 
 To generate the HTML User Guide, you will need:
 
-- `Sphinx <http://www.sphinx-doc.org/en/master/>`_ of compatible version,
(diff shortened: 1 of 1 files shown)


hjoliver:
> (e.g. the functional tests which use non-POSIX stuff like sed -r)

Approving, but note that the functional tests are not going to be distributed with Cylc releases. So only developers working with Cylc git clones need to care about non coreutils commands used in tests.

PR #95, opened 2019-10-23
fail build on invalid references

oliver-sanders:
Cause the Sphinx build to fail if any references are missing / corrupted. E.G. if you try to reference a python function which isn't documented you get an error.

This will be especially useful when we document more things like Cylc configurations etc.

This should fail until https://github.com/cylc/cylc-flow/pull/3417 is merged.

### DIFF
--- a/src/conf.py
+++ b/src/conf.py
@@ -86,6 +86,9 @@ release = CYLC_VERSION  # The full version, including alpha/beta/rc tags.
 intersphinx_mapping = {
     'rose': (
         'http://metomi.github.io/rose/doc/html', None
+    ),
+    'python': (
+        'https://docs.python.org/', None
     )
 }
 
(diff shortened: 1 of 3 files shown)


src/conf.py
```diff
@@ -86,6 +86,9 @@
 intersphinx_mapping = {
     'rose': (
         'http://metomi.github.io/rose/doc/html', None
+    ),
+    'python': (
```
oliver-sanders: The Python reference material defines things like `PYTHONPATH` which is why I've turned it on in inter-sphinx.

src/tutorial/runtime/introduction.rst
```diff
@@ -69,16 +69,16 @@ We can also call other scripts or executables in this way, e.g:
            script = ~/foo/bar/baz/hello_world
 
 
-:envvar:`PATH` and :envvar:`PYTHONPATH`
----------------------------------------
+``PATH`` and :envvar:`PYTHONPATH`
```
oliver-sanders: The Python reference material does not document the `PATH` variable and I can't find anywhere appropriate to source it from so the simple solution is to demote it from a reference to a string literal.

PR #96, opened 2019-10-30
Document no message send retries.

### LINKED ISSUE #87: Document why task messages are not retried in Cylc 8
See https://github.com/cylc/cylc-flow/pull/3348#discussion_r332341533

hjoliver:
Close #87 

One review will do @dpmatthews

### DIFF
--- a/src/task-implementation.rst
+++ b/src/task-implementation.rst
@@ -50,12 +50,16 @@ independent testing, re-use, and shell mode editing.
 Task Messages
 -------------
 
-Task jobs automatically send status messages back to the server program to
-report that execution has started, succeeded, or failed. Custom messages
-can also be sent back by the same mechanism, with various severity levels.
-These can be used to trigger other tasks off specific task outputs, or 
-to trigger execution of event handlers by the suite server program (see
-:ref:`EventHandling`), or just to write information to the server log.
+Task jobs send status messages back to the server program to report that
+execution has started, succeeded, or failed. Custom messages can also be sent
+by the same mechanism, with various severity levels.  These can be used to
+trigger other tasks off specific task outputs, or to trigger execution of event
+handlers by the server program (see :ref:`EventHandling`), or just to write
+information to the server log.
+
+(If polling is configured as the task communication method for a host, the
+messaging system just writes messages to the local job status file for
+recovery by the server at the next poll).
 
 Normal severity messages are printed to ``job.out`` and logged by the
 server program:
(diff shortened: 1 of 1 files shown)


src/task-implementation.rst
```diff
@@ -50,22 +50,22 @@ independent testing, re-use, and shell mode editing.
...
+Task jobs automatically send status messages back to the server program to
+report that execution has started, succeeded, or failed. Custom messages
+can also be sent back by the same mechanism, with various severity levels.
+These can be used to trigger other tasks off specific task outputs, or 
+to trigger execution of event handlers by the suite server program (see
+:ref:`EventHandling`), or just to write information to the server log.
 
 Normal severity messages are printed to ``job.out`` and logged by the
-suite server program:
+server program:
```
dpmatthews: "suite" removed here but not in multiple places below. Probably not worth worrying about - we'll need a major review of the docs if we want to get rid of all references to "suite server".

src/task-implementation.rst
```diff
@@ -74,27 +74,31 @@ event handlers:
...
 .. code-block:: bash
 
    cylc message -- "${CYLC_SUITE_NAME}" "${CYLC_TASK_JOB}" \
      "CRITICAL:ERROR occurred in process X!"
 
+Task jobs no longer (since Cylc 8) attempt to resend messages if the server
+cannot be reached. Send failures normally imply a network or Cylc configuration
+problem that will not recover by itself, in which case a series of messaging
+retries just holds up job completion unnecessarily. The server will recover
+task status by polling on job timeout, if a message does not get through.
```
dpmatthews: Perhaps add
... on job timeout (or earlier if execution polling is configured)

src/task-implementation.rst
```diff
@@ -50,22 +50,22 @@ independent testing, re-use, and shell mode editing.
 Task Messages
 -------------
 
-Tasks messages can be sent back to the suite server program to report completed
-outputs and arbitrary messages of different severity levels.
-
-Some types of message - in addition to events like task failure -  can
-optionally trigger execution of event handlers in the suite server program
-(see :ref:`EventHandling`).
+Task jobs automatically send status messages back to the server program to
```
dpmatthews: Clarify that this doesn't happen if you set task communication method = poll ?

src/task-implementation.rst
hjoliver: Yeah, I'm trying to avoid using "suite" in new documentation updates, but fairly half-heartedly because I'm aware we'll have to do a full sweep through all the docs before the 8.0.0 release.

src/task-implementation.rst
hjoliver: Done.

src/task-implementation.rst
hjoliver: Done.

hjoliver:
Feedback addressed.

PR #98, opened 2019-11-04
fix external links

oliver-sanders:


src/appendices/site-user-config-ref.rst
```diff
@@ -136,7 +136,6 @@ URL of the cylc internet homepage, with links to documentation for the
 latest official release.
 
 - *type*: string
-- *default*: https://cylc.github.com/cylc-flow/
```
oliver-sanders: Configuration docs are drifting from reality and will need a major overhaul after current work is completed so happy to ditch this for now.

PR #102, opened 2019-11-20
jinja2: cylc logging

oliver-sanders:
*Result of recent user support*.

Cylc8 test PR: https://github.com/cylc/cylc-flow/pull/3448

Users can gain access to the Cylc logger from within Jinja2.

Should we encourage and document this?

### DIFF
--- a/src/suite-config.rst
+++ b/src/suite-config.rst
@@ -3519,6 +3519,38 @@ For better clarity and disambiguation Python modules can be prefixed with
    {% from "__python__.itertools" import product %}
 
 
+Logging
+^^^^^^^
+
+It is possible to output messages to the Cylc log from within Jinja2, these
+messages will appear on the console when running or validating a suite.
+This can be useful for development or debugging.
+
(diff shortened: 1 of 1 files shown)


kinow:
>This can be useful for development or debugging.

+1 It sounds really useful for development & debugging as you pointed in the doc. Probably risky to use it in production, especially if parsing/monitoring the logs. I will probably end up using it for debugging issues now that I know it's possible haha.

oliver-sanders:
>  Probably risky to use it in production, especially if parsing/monitoring the logs

I don't think users do suite log parsing but agreed it could be kinda nasty. This will enevitably be used for printing log messages from within loops, namely task generation. They could end up with many thousands of lines of log output.

We could add the advice to use debug level logging i.e:

```
{% do LOG.debug("Hello World!") %}
```

```console
$ cylc validate --debug <suite>
```

oliver-sanders:
OK, will update advice to use the debug level and add a test to Cylc Flow.

PR #103, opened 2019-11-24
Fix example in parameter expansion

kinow:
Three commits, one uses `1..5` for parameter expansion. The examples appear to be related to a previous paragraph, where the docs use `1..5`, `ship, buoy, plane` for obs, etc. The examples match the content in that previous paragraph, except for this one, that is using 1..2.

The other commits were to align the multiline string in the examples. There is one example with multiline string that is aligned correctly, but the two subsequent examples appear to be off by a couple spaces?

### DIFF
--- a/src/suite-config.rst
+++ b/src/suite-config.rst
@@ -2787,7 +2787,7 @@ expands to:
 .. code-block:: cylc
 
    R1 = """proc_small => proc_big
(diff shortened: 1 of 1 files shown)


hjoliver:
@kinow - can you repeat this one for 7.8.5 docs? (it looks like we will make another 7.8.x release).

PR #108, opened 2020-01-12
Execute user scripts in a subshell - documentation

TomekTrzeciak:
Documentation update for cylc/cylc-flow#3440.

~Note: it would be worth to update `img/anatomy-of-a-job-script.png` as well, but I don't have a source for it.~


hjoliver:
> Note: it would be worth to update img/anatomy-of-a-job-script.png as well, but I don't have a source for it.

Just to note, before we merge this, after the code PR is merged - I can do this diagram update.

TomekTrzeciak:
Mentioned [here](https://cylc.discourse.group/t/separating-task-stdout-from-stdout-of-script-within-buffering-issue/210/8), `wait $CYLC_TASK_MESSAGE_STARTED_PID` in a user script appears be [officially documented](https://cylc.github.io/doc/built-sphinx-single/index.html#checkpointing-with-a-task). The associated code PR cylc/cylc-flow#3440 will require replacing this with `cylc__job__wait_cylc_message_started` shell function, but perhaps a more decent name is called for if this is going to be an official feature. @hjoliver, any opinion?

hjoliver:
Thanks for flagging this @TomekTrzeciak. I think it does need to be an official feature, and that function name seems reasonable to me :+1:

TomekTrzeciak:
> Thanks for flagging this @TomekTrzeciak. I think it does need to be an official feature, and that function name seems reasonable to me 👍

OK, I updated the docs. Still, it feels like this background message started process is an implementation detail that the user doesn't need to know or care about. Perhaps it would be better to have some cylc commands to be aware of the background message process (e.g. by inspecting CYLC_TASK_MESSAGE_STARTED_PID variable) and wait for it if necessary. This way the race condition with custom message triggers could be avoided as well, cf.  https://github.com/cylc/cylc-flow/pull/3440#issuecomment-557118889. Also, the implementation could be changed without affecting user scripts.

hjoliver:
> Still, it feels like this background message started process is an implementation detail that the user doesn't need to know or care about

That's generally true, except when users do not want message stdout interleaved with their job stdout (e.g. they need to parse job stdout). 

> Perhaps it would be better to have some cylc commands to be aware of the background message process (e.g. by inspecting CYLC_TASK_MESSAGE_STARTED_PID variable) and wait for it if necessary.

Do you mean a special command for that purpose? (presumably you do, as users typically do not use existing Cylc commands at the start of their job script). But if so, how is that any different (as far as users are concerned) from your shell function?  (Just that the function name makes it look like "low level" functionality?).

TomekTrzeciak:
> Do you mean a special command for that purpose?

No, I actually meant commands like cylc message and checkpoint and any others that might be sensitive to `waiting -> running` state change. They could check and wait for the pending message started that will transition the task state. This still wouldn't be 100% reliable (message started could fail, while the checkpoint or custom message could succeed), but at least it would avoid the race condition in most cases.

Most users won't be knowledgeable enough to implement their scripts to avoid this kind of problems. They will just test that they work reliably, which they may under normal conditions, but could start failing unexpectedly under e.g. high network congestion or other problems that may lead to unpredictable order of network communication.

hjoliver:
Ah, got it. I thought you were also suggesting this as a way for users to avoid interleaved message output.  But we probably should do as you say for the potential race condition.

TomekTrzeciak:
Updated to the latest master, including the diagram (thanks @oliver-sanders for making it editable 👍 ).

(the build fails with `src/installation.rst:15:Unexpected indentation`, also on master).

hjoliver (approved):
:+1:

PR #105, opened 2019-11-29
Adding Message Trigger Tutorial to cylc-docs.

datamel:
Changed slightly from the original rose commit to clarify terminology.

PR #112, opened 2020-02-06
Minor doc tweak.

hjoliver:
Corrects a backwards use of the word "trigger" in the glossary. Assigning @datamel as I found this when reading beyond the bounds of #105.  One review will do.

### DIFF
--- a/src/glossary.rst
+++ b/src/glossary.rst
@@ -92,10 +92,11 @@ Glossary
       :term:`Dependency <dependency>` relationships can be thought of the other
       way around as "triggers".
 
-      For example the dependency ``foo => bar`` could be described in two ways:
+      For example the dependency ``foo => bar`` could be described in several ways:
 
-      * "``bar`` is dependent on ``foo``"
-      * "``foo`` is triggered by ``bar``"
+      * "``bar`` depends on ``foo``"
+      * "``foo`` triggers ``bar``"
+      * "``bar`` triggers off of ``foo``"
 
       In practice a trigger is the left-hand side of a dependency (``foo`` in
       this example).


src/glossary.rst
```diff
@@ -92,10 +92,11 @@ Glossary
...
       way around as "triggers".
 
-      For example the dependency ``foo => bar`` could be described in two ways:
+      For example the dependency ``foo => bar`` could be described in several ways:
 
-      * "``bar`` is dependent on ``foo``"
-      * "``foo`` is triggered by ``bar``"
+      * "``bar`` depends on ``foo``"
+      * "``foo`` triggers ``bar``"
+      * "``bar`` triggers off of ``foo``"
```
hjoliver: Using the American "off of" here because it is unambigous (The non-American "x triggers off y" implies the wrong direction to some people - like "x kicks off y").

datamel: I agree that 'triggers off of' is less ambiguous than 'triggers off'. A bit of a mouthful but can't think of anything better if wanting to move away from 'is triggered by'.

hjoliver: Well "`bar` is triggered by `foo`" is fine too, but the original line has it the wrong way around. 

Actually, "`bar` is triggered by `foo`" implies the upstream task `foo` is the protagonist,  whereas "`bar` triggers off of `foo`" suggests that the downstream task `bar` triggers itself in response to the state of `foo`.

Under spawn on submit, it is actually the latter case that is correct (to the extent that it matters) ... but in our spawn on demand future, the former case will be true.  

(Hopefully that's clear as mud!)

PR #115, opened 2020-02-21
Add Conda installation instructions

kinow:
From the `README.md` file of https://github.com/cylc/cylc-conda/. When merged, the `cylc/cylc-conda` repository should be ready to be archived.

@hjoliver tried to keep as much as possible from the content in the README. Trying my best to convert Markdown to reStructuredText, using [this as guideline](https://docutils.sourceforge.io/docs/user/rst/quickref.html) as well as looking at existing content in the docs.

Happy to update if necessary, these are the main points that I think need some attention for review:

- Location of this docs? I've put this under Installation. But maybe it should go under Installing Cylc? 
- Images. I can't recall exactly, but I think there was some step to copy or resize images for PDF? Memory is blur (almost home-time on Friday), but just in case that's still important.
- Images #2. I've added the images under `img/installation/conda/`... not sure if that's a problem? Otherwise I feel like the `img` folder will get bloated. Not sure if that's OK or not?
- I've added a `(cylc1) $` to show that those lines need the environment activated. Happy to change that if necessary.


kinow:
And to review, here's one way to build it locally (I used `pip`, `conda` should work too).

1. clone
2. new venv
3. activate
4. `pip install -e .[all]`
5. `cd ../cylc-flow/ && pip install -e .` (assuming `master`, latest)
6. `cd -; make clean html`

Then open the `index.html` from `doc/8.0a1/` in a browser.

image

src/installation.rst
```diff
@@ -157,3 +157,154 @@ Cylc commands. Task job scripts invoke ``bash -l`` (login shells) to run the
...
+
+Enable the "kinow" Conda channel in your ``~/.condarc`` file
+(pending package availability on conda-forge).
+
+::
+
+   channels:
+     - defaults
+     - conda-forge
+     - kinow
```
kinow: This is not necessary any longer! 🎉

kinow: (removed)

PR #117, opened 2020-02-26
Add documentation for the WebSockets used in Cylc UI

kinow:
Not sure if "Architecture" is the best section for that. Actually, perhaps the documentation is too much developer-oriented. If so, happy to close and move it to a Wiki somewhere, another section, etc.

image


kinow:
Hmm, restarted the build a couple times, but it keeps failing. It appears to be some `cylc validate` command somewhere that is failing. Tried the same command locally and it fails indeed. Removed cache in Travis, still fails.

Had a look at the logs, seems to be a problem with a task in a suite. The CI runs a script `validate` script in some folders. In one of the suites, `cylc-forecasting-suite`, the `validate` script will run the suite (others run just `cylc validate`).

The `cylc-forecasting-suite` has tasks that execute [`get-observations`](https://github.com/cylc/cylc-doc/blob/a65f52339c4f255f804350969900886b8b9317a4/cylc/doc/etc/tutorial/cylc-forecasting-suite/bin/get-observations). This script calls "http://datapoint.metoffice.gov.uk/public/data/", which appears to be returning an error code, breaking the build.

It should be fixed eventually when that endpoint gets back to normal :+1: not related to this PR I think.

image

datamel:
I'm sure people know more about the ins and outs of this than I do but thought I would take a look! I have read through this and in my opinion it is well written and informative. 
Yes, the error message is not to do with the PR (and appearing throughout other prs and master).
My only very minor suggestion, if this is intended for users (I agree it being more developer-oriented), is perhaps make a reference to the `ack` being an acknowlegdement, although I'm happy to be ignored!

kinow:
Hi Mel,

Thanks for the review! Good idea to clarify on the `ack` message. Here's what I've changed, including a note that the protocol says it MAY have an ack, not MUST, so that message isn't always sent back.

I learned it when we were investigating issues with WebSockets, and the `ack` would either not be sent to the client, or sent out of order. In both cases, the JS client (ApolloClient) worked fine, as I think it doesn't care much about this `ack` message.

```diff
diff --git a/src/architecture.rst b/src/architecture.rst
index 4a74100..c28cfe3 100644
--- a/src/architecture.rst
+++ b/src/architecture.rst
@@ -38,7 +38,11 @@ called `graphql-ws protocol`_.
 
 After a channel between client and server is open, the messages
 follow that protocol, starting by the ``connection init`` message,
-that simply expects an ``ack`` back from the server.
+that simply expects an ``ack`` message back from the server,
+where the ``ack`` is an acknowlegdement to the client - note
+that the protocol does not define an ``ack`` as a MUST, but
+rather as a MAY, so a client may interpret not receiving an
+error as an acknowledgement to proceed as well.
 
 The next message will be a ``start``, which will contain the
 GraphQL query subscription. If there were no errors, the client and
```

datamel:
> Hi Mel,
> Thanks for the review! Good idea to clarify on the ack message. Here's what I've changed, including a note that the protocol says it MAY have an ack, not MUST, so that message isn't always sent back.
> I learned it when we were investigating issues with WebSockets, and the ack would either not be sent to the client, or sent out of order. In both cases, the JS client (ApolloClient) worked fine, as I think it doesn't care much about this ack message.
> diff --git a/src/architecture.rst b/src/architecture.rst
> index 4a74100..c28cfe3 100644
> --- a/src/architecture.rst
> +++ b/src/architecture.rst
> @@ -38,7 +38,11 @@ called `graphql-ws protocol`_.
>  
>  After a channel between client and server is open, the messages
>  follow that protocol, starting by the ``connection init`` message,
> -that simply expects an ``ack`` back from the server.
> +that simply expects an ``ack`` message back from the server,
> +where the ``ack`` is an acknowlegdement to the client - note
> +that the protocol does not define an ``ack`` as a MUST, but
> +rather as a MAY, so a client may interpret not receiving an
> +error as an acknowledgement to proceed as well.
>  
>  The next message will be a ``start``, which will contain the
>  GraphQL query subscription. If there were no errors, the client and
> WDYT?

Great stuff @kinow, nice to have a really clear explanation!

oliver-sanders:
@kinow I can't restart Travis atm, if you rebase & push that might clear it.

oliver-sanders:
(note you may need to ignore TB failures on the tutorial suite for now)

PR #121, opened 2020-04-29
Conf tree

oliver-sanders:
**Sibling PRs:**

* https://github.com/cylc/cylc-flow/pull/3559
* https://github.com/cylc/cylc-sphinx-extensions/pull/22

**Highlights:**

* Use the cylc-sphinx-extensions library (remove old built-in extensions).
* Autodocument the Cylc suite and global configurations.
* Fix broken references.

**Note:**

The presentation is down to the Sphinx theme not the documentation method.

**Further Work:**

* For configs shared between the suite and global configs, I've only documented one copy. Once the platforms work is done I'll work out a way of automatically referencing one from the other.
* The docs still require reviewing, I've just copied them from a to b and fixed things up a little.
* I've only corrected the broken references for now, ultimately we will need to go through the code making changes like this:

    ```diff
     - See the ``[scheduling]`` section.
     + See the :cylc:conf:`[scheduling]` section.
    ```

**Benefits:**

* Consistency with Cylc Flow, obvious.
* Integrity checking, the docs will fail to build if a configuration is removed, but still referenced in the docs.
* Links, all references to cylc sections result in hyperlinks to the relevant content.
* Collaboration, objects can be shared with other projects (e.g. we can reference Cylc configurations from the Rose docs).

**Screenshots**

Screenshot 2020-04-29 at 11 51 42
Screenshot 2020-04-29 at 11 52 15
Screenshot 2020-04-29 at 11 52 46

PR #122, opened 2020-04-29
decouple sphinx version

oliver-sanders:
The rate of Sphinx development is too fast to keep manually pinning it to a subversion ~(we will soon be on Sphinx3)~ Sphinx 3 is out.

Also we are using tools, which for good reason do pin the sphinx version. So at present Cylc-doc is incompatible with cylc-sphinx-extensions.

Set a looser definition of what we want based on what we actually need.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,6 @@
 # More information for Travis can be found at http://docs.travis-ci.com/user/getting-started/
 
 ---
-cache: pip
 dist: xenial
 
 language: python
--- a/setup.py
+++ b/setup.py
@@ -39,7 +39,7 @@ VERSION = get_version(
 )
 
 INSTALL_REQUIRES = [
-    'sphinx==2.0.*',
+    'sphinx>=2.0.0,<3.0.0',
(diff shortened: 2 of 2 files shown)


oliver-sanders:
Waiting on https://github.com/cylc/cylc-sphinx-extensions/pull/24

oliver-sanders:
Now failing for a different reason!

Can't locate the `cylc.main_loop.health_check` plugin, hmmm, ...

kinow:
> Now failing for a different reason!
> 
> Can't locate the `cylc.main_loop.health_check` plugin, hmmm, ...

Not sure if helpful, but I had the exact same issue with a docker container running those multiple bash versions. I think I switched branches, then had to `pip install -e .` again, and everything worked. So it could be something cached somewhere (you can purge caches in Travis if necessary). Just in case it helps.

oliver-sanders:
Good idea, would have made sense, sadly didn't work. Might leave it for the weekend, then delete caches and try again Monday.

oliver-sanders:
Stuff it, I'll bundle this with #121

PR #125, opened 2020-05-04
travis => actions

oliver-sanders:
Also includes:

* sphinx version change from #122.
* tutorial fixes ported from https://github.com/metomi/rose/pull/2398.
* PIL import fix.
* converted todo directives to comments.

See it in action on my fork: https://github.com/oliver-sanders/cylc-doc/pull/2


.github/workflows/test.yml
```diff
@@ -0,0 +1,65 @@
+name: test
+
+on: [pull_request]
```
wxtim, line 3: Can we configure it to run periodically (weekly?) on master just to double check that master hasn't broken?

MetRonnie: Yes that's possible: https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#onschedule though I'm curious as to how it would break spontaneously?

MetRonnie: Actually I suppose stuff like the API used in the tutorial might break

oliver-sanders: Yes we can, though it becomes a bit of a tree falling in the forrest problem. If the action fails but no-one bothers to look, did it fail?

This issue applies to all of our repos (due to inter-dependence) so I think we will require a more integrated solution, there is a an issue open somewhere (I think on cylc admin) to setup nightly builds to catch these sorts of problems.

wxtim: Let's not worry about this then...

PR #126, opened 2020-05-13
fixed badge

wxtim:
Fix badge in readme referring to Travis.

I've tried to follow the instructions at https://help.github.com/en/actions/configuring-and-managing-workflows/configuring-a-workflow#adding-a-workflow-status-badge-to-your-repository but can't get it working...

@MetRonnie @oliver-sanders - any thoughts?>

PR #127, opened 2020-05-14
Fix badge

MetRonnie:
- The test workflow now also runs when a commit is pushed to master (including PR merge).
- The badge only shows the status of test workflow run on push on master

This supersedes #126


MetRonnie:
Yes, all of that is correct. It's the only way to show the test status of master

PR #129, opened 2020-05-21
installation: re-write

oliver-sanders:
Jot down the Mac OS installation instructions whilst they are fresh-ish in my head (will get re-jigged when we setup Mac OS testing on GH Actions).

Had a go at the installation instructions whilst I was at it. It's a first pass which will require more work but for the moment it's probably sufficient for alpha2.

### DIFF
--- a/src/installation.rst
+++ b/src/installation.rst
@@ -118,6 +118,16 @@ your ``$PATH``, follow the instructions in the ``brew install`` output.
                        self._host_exs[target] = socket.gethostbyname_ex(target)
                    except IOError as exc:
 
+.. warning::
+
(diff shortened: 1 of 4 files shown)


src/installation.rst
```diff
@@ -1,301 +1,288 @@
...
-.. _GNU Coreutils: https://www.gnu.org/software/coreutils/coreutils.html
+   It has a fully functional Python 3 workflow service and CLI that can run
+   existing Cylc workflows.
 
-Requirements:
+   **BUT:**
 
-- Python 3.7+
+   - It is not production-ready yet
+     - Use the latest cylc-7.8 release in production
```
kinow, line 14: Or 7.9 if Py2.6+?

PR #133, opened 2020-07-16
actions: automatically [un]deploy documented versions

oliver-sanders:
**After two previous aborted attempts to do this with Rose we finally have a system which ticks the required automation and security boxes:**

* Two GitHub actions called `deploy` and `undeploy`.
* They must be manually triggered by users with write permissions.
* These actions add new commits to the `gh-pages` branch (which allows easy rollback).

More info in `README.md`.

**Under the hood the `deploy` action:**

* Checks out cylc-doc and gh-pages, then symlinks the `doc/` dir from cylc-doc to point at `gh-pages`.
* That way the Makefile builds to docs directly into the gh-pages branch.
* This means we use the same logic for building locally as we do for deploying right down to the multi-version support.

**KISS:**

Everything can still be done manually, and this is useful as it enables us to manually install the Cylc7 docs into the repo (just `cp -r` the docs across and update the `versions.json` file accordingly).

**Tags?**

The deploy action:

* *Checks out* cylc-doc at the provided tag.
* *`pip installs`* cylc-flow at the provided version.

The cylc-flow version (which is the same as the cylc-flow tag) must match the cylc-doc tag (otherwise nothing will happen). A little funky but quite purposeful.

Note that the docs are built with the cylc-doc scripts *from the repository at that version*, this is good, it means we can safely re-build older docs (unlike other sphinx deploy systems which rebuild all versions whenever you deploy anything) but it does mean I've had to forward port some changes from this branch into the `8.0a2` tag which is already on this repo.

**Proof that it works:**

* https://cylc.github.io/cylc-doc/current/html/index.html
* https://github.com/oliver-sanders/cylc-doc/runs/878320584?check_suite_focus=true


.github/workflows/deploy.yml
```diff
@@ -99,11 +99,10 @@ jobs:
             clean \
             html \
             slides \
+            linkcheck \
```
oliver-sanders, line 102: To add to the fun there is a broken URL on cylc-flow (which we failed to spot) and that causes deployment to fail (and rightly so).

It's not worth fiddling the cylc-flow 8.0a2 tag (and I don't think we would do that anyway) so I removed `linkcheck` from the build to get the deployment running.

If linkcheck becomes problematic in the future we may need to find other ways to cope with it, but putting `linkcheck` back in for now.

bin/version
```diff
@@ -0,0 +1,137 @@
...
+
+def tidy(args):
+    # get list of documented versions
+    versions = [
+        parse_version(version)
+        for version in get_formats(args.builddir)
+    ]
+
+    # build dictionary of minor versions
+    # e.g. {(1, 0): ['1.0.1', '1.0.2', '1.0.3']}
```
wxtim, line 70:
```suggestion
    # versions = [<Version('1.0.1')>, <Version('1.0.2')>]
    # minor_versions = {
    #     (1, 0): [<Version('1.0.1')>, <Version('1.0.2')>]
    # }
```

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,125 @@
...
+          )
+
+      - name: push changes
+        run: |
+          (
+            cd gh-pages
+            git add *
+            git commit -m "add: ${{ github.event.inputs.tag }}"
+            git push origin HEAD
+          )
```
MetRonnie, line 125: Out of curiosity, why run these commands in a subshell?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,125 @@
+# THIS FILE IS PART OF THE CYLC SUITE ENGINE.
+# Copyright (C) 2008-2019 NIWA & British Crown (Met Office) & Contributors.
```
MetRonnie, line 2: Minor niggle - copyright year out of date

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,125 @@
...
+      - name: configure git
+        run: |
+          git config --global user.name "action:deploy"
+          git config --global user.email "action:deploy@github.com"
+
+      - name: tidy old versions
+        run: |
+          (
+            cd docs
+            git rm -r $("${{ github.workspace }}/docs/bin/version" tidy) || true
```
MetRonnie, line 115: [From the demo run:](https://github.com/oliver-sanders/cylc-doc/runs/878320584?check_suite_focus=true)
> ```
> line 3: /home/runner/work/cylc-doc/cylc-doc/docs/bin/version: No such file or directory
> fatal: No pathspec was given. Which files should I remove?
> ```
Is this expected?

.github/workflows/deploy.yml
oliver-sanders: This is a habit of mine to wrap logic which `cd`s places in subshells, it might not be necessary in GH actions?

.github/workflows/deploy.yml
oliver-sanders: Well spotted.

.github/workflows/deploy.yml
oliver-sanders: Yes, ugly I know but it's the natural result of `git rm -r  # nothing`.

.github/workflows/deploy.yml
MetRonnie: Oh, yeah each step runs in individual shells

PR #135, opened 2020-07-23
actions: nightly builds

oliver-sanders:
Perform a nightly build of `cylc-doc:master` against `cylc-flow:master` at around 0:35UTC.

I've managed to jigger up GH actions on my own fork which is no longer recognising the `nightly` action so I've kinda deleted the evidence that this works, will try and get it going again, in the mean time take a look at my `gh-pages` branch:

https://oliver-sanders.github.io/cylc-doc/current/html/index.html

When we perform a new nightly build we delete the commit of the last one so that the branch history doesn't inflate massively.

Sadly this is a separate workflow from `deploy` as there is just too much different between the two to unify them in any meaningful way.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -86,7 +86,7 @@ jobs:
 
       - name: remove previous nightly build
         run: |
-          EDITOR='sed -i "s/pick \([0-9a-z]* -nightly build-\)/omit \1/"' \
+          EDITOR='sed -i "s/pick \([0-9a-z]* -nightly build-\)/drop \1/"' \
             git -C gh-pages rebase -i --keep-base
 
       - name: build docs
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -75,44 +75,33 @@ jobs:
           DOCS="${{ github.workspace }}/docs" \
           PAGE="${{ github.workspace }}/gh-pages"  \
 
-          (
-            cd "$DOCS"
-            rm -r doc
-            ln -s "$PAGE" doc
-          )
-
-          ls -l "$DOCS"
-
-      - name: build docs
-        run: |
-          # remove previous build
-          rm -r docs/nightly || true
-          # make new build
-          make -C docs \
-            html \
-            slides \
-            linkcheck \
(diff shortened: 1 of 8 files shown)


.github/workflows/nightly.yml
```diff
@@ -0,0 +1,107 @@
...
+
+# Deploy a new version of the documentation or rebuild an existing one.
+
+# Note: All changes made to the gh-pages branch are non-destructive 
+#       (i.e. no force pushing) so all changes can be undone.
+
+name: nightly
+
+on:
+  workflow_dispatch:
```
oliver-sanders, line 25: This means we can manually trigger the action.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,107 @@
...
+          ln -s "$PAGE" doc
+
+      - name: configure git
+        run: |
+          git config --global user.name "action:deploy"
+          git config --global user.email "action:deploy@github.com"
+
+      - name: remove previous nightly build
+        run: |
+          EDITOR='sed -i "s/pick \([0-9a-z]* -nightly build-\)/omit \1/"' \
```
oliver-sanders, line 89: Use `sed` as the text editor in order to delete the offending commit via rebase.

MetRonnie: Shouldn't it be `drop` instead of `omit`?

MetRonnie:
Minor thing, the versions get a bit "chopped up" on the webpage:

image

This could be fixed by settting the anchor elements' CSS to `display: inline-block`, though I don't know if that's something we have control over

.github/workflows/nightly.yml
oliver-sanders: Yes, I think omit might have been a placeholder I failed to fix.

oliver-sanders:
We do have control over it, however, it's a bit of a pain, there will only ever be one nightly build so probably not an issue.

oliver-sanders:
Let's stick this in and see what happens, it's not the live docs anyway.

oliver-sanders:
https://github.com/cylc/cylc-doc/actions/runs/185468390

PR #139, opened 2020-08-03
Fix nightly build workflow

MetRonnie:
Nightly builds were supposed to be dropped from commit history, but weren't.

Problem was `git rebase -i --keep-base` wasn't getting any commits. Also need to force push.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -16,8 +16,8 @@
 
 # Deploy a new version of the documentation or rebuild an existing one.
 
-# Note: All changes made to the gh-pages branch are non-destructive 
-#       (i.e. no force pushing) so all changes can be undone.
+# Note: This action deletes its previous commits so the nightly build history
+# is not preserved and does not require housekeeping.
 
 name: nightly
 
@@ -36,7 +36,7 @@ jobs:
           python-version: '3.7'
 
       - name: checkout cylc-doc
-        uses: actions/checkout@v2
+        uses: actions/checkout@v2.3.1
         with:
           ref: master
           path: docs
@@ -52,7 +52,7 @@ jobs:
           pip install "${{ github.workspace }}/docs[all]"
 
       - name: checkout cylc-flow
-        uses: actions/checkout@v2
+        uses: actions/checkout@v2.3.1
         with:
           repository: cylc/cylc-flow
           ref: master
@@ -64,11 +64,12 @@ jobs:
           #       have extra dependencies.
           pip install './cylc-flow[all]'
 
-      - name: checkout gh-pages
-        uses: actions/checkout@v2
+      - name: checkout gh-pages branch
+        uses: actions/checkout@v2.3.1
         with:
           ref: gh-pages
           path: gh-pages
+          fetch-depth: 0 # all history (unfortunately all branches too)
 
       - name: install gh-pages
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly.yml
```diff
@@ -64,11 +64,12 @@ jobs:
...
           pip install './cylc-flow[all]'
 
-      - name: checkout gh-pages
-        uses: actions/checkout@v2
+      - name: checkout gh-pages branch
+        uses: actions/checkout@v2.3.1
         with:
           ref: gh-pages
           path: gh-pages
+          fetch-depth: 0 # all history (unfortunately all branches too)
```
oliver-sanders, line 72: Derp, of course it was only ever working with a single commit before :/

.github/workflows/nightly.yml
```diff
@@ -81,17 +82,20 @@ jobs:
...
 
       - name: remove previous nightly build
         run: |
           cd gh-pages
+          echo "INFO: gh-pages branch at $(git rev-parse HEAD) pre-rebase"
+          echo "Use git fetch <upstream> <commit> if you need to recover the branch"
+          echo ""
           EDITOR='sed -i "s/pick \([0-9a-z]* -nightly build-\)/drop \1/"' \
-            git rebase -i --keep-base
+            git rebase -i gh-pages.root
```
oliver-sanders, line 95: This doesn't work for me, how about `git rebase -i --root`?

oliver-sanders:
Removed the surplus nightly builds to give this action a chance of passing when it's merged.

MetRonnie:
> Removed the surplus nightly builds to give this action a chance of passing when it's merged.

Actually the rebase in the workflow would have taken care of all of them anyway

.github/workflows/nightly.yml
MetRonnie: Ah, should be `git rebase -i origin/gh-pages.root` (I created that branch)

.github/workflows/nightly.yml
```diff
@@ -81,17 +82,20 @@ jobs:
...
 
       - name: remove previous nightly build
         run: |
           cd gh-pages
+          echo "INFO: gh-pages branch at $(git rev-parse HEAD) pre-rebase"
+          echo "Use git fetch <upstream> <commit> if you need to recover the branch"
+          echo ""
           EDITOR='sed -i "s/pick \([0-9a-z]* -nightly build-\)/drop \1/"' \
-            git rebase -i --keep-base
+            git rebase -i origin/gh-pages.root
```
oliver-sanders, line 95: I'm not convinced by this `gh-pages.root` approach, it's a potential breakage.

MetRonnie: Ok, then just use `--root`?

oliver-sanders: Fine by me, probs not worth trying to work out the previous commit hash for rebase.

oliver-sanders:
Test failure is due to the (ever flaky) tutorial suite so can be ignored.

.github/workflows/nightly.yml
```diff
@@ -109,6 +113,8 @@ jobs:
           git -C gh-pages add "$build_name" 'versions.json'
 
       - name: push changes
+        working-directory: gh-pages
         run: |
-          git -C gh-pages commit -m '-nightly build-'
-          git -C gh-pages push origin HEAD
+          git status
+          git commit -m '-nightly build-'
+          git push --force-with-lease origin gh-pages
```
oliver-sanders, line 120: `--force-with-lease`, had to google that one.

oliver-sanders:
Happy to slide this one in with one approval as it is a nightly build of an undeployed system, let's see it work.

PR #140, opened 2020-08-04
validate xtrigger labels

oliver-sanders:
Companion to https://github.com/cylc/cylc-flow/pull/3732.


oliver-sanders:
Tests will pass when the companion PR is merged.

PR #143, opened 2020-08-10
introduction

oliver-sanders:
Tidy up the top-level sections of the docs by creating a new "introduction" section.


oliver-sanders:
No content change and tests pass so one review should be sufficient.

PR #150, opened 2020-08-19
setup: up sphinx_rtd_theme version

oliver-sanders:
The sphinx_rtd_theme project either had a bug or made a bad design decision (not sure which) over the styling of definition lists, as of 0.5.0 it has been fixed/reverted.

This PR pins the version so we don't get that nasty styling again:

**Before:**

Screenshot 2020-08-19 at 17 26 09

**After:**

Screenshot 2020-08-19 at 17 26 18

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,7 @@ VERSION = get_version(
 
 INSTALL_REQUIRES = [
     'sphinx>=2.0.0,<3.0.0',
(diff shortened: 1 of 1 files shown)


wxtim (approved):
It's broken in the same way as master, demanding fix of master out of scope of PR.

PR #151, opened 2020-08-19
hieroglyph_patch: remove obsolete extension

oliver-sanders:
Sibling PR: https://github.com/cylc/cylc-sphinx-extensions/pull/32

The `hieroglyph_patch` extension is a dirty hack to get around this issue:

https://github.com/nyergler/hieroglyph/issues/148

But now that hieroglyph 2.1.0 has been released we no longer need it!

This removes the Sphinx version coupling which was locking us into an ever decreasingly small number of compatible Sphinx2 versions.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,12 +39,12 @@ VERSION = get_version(
 )
 
 INSTALL_REQUIRES = [
-    'sphinx>=2.0.0,<3.0.0',
+    'cylc-sphinx-extensions>=1.2'
+    'eralchemy==1.2.*',
+    'hieroglyph>=2.1.0',
+    'sphinx>=3.0.0',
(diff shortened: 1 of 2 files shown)


kinow (approved):

oliver-sanders:
(tests broken by suite.rc -> flow.cylc change)

kinow:
> (tests broken by suite.rc -> flow.cylc change)

:+1:   I thought they would be broken due to the other hieroglyph PR

oliver-sanders:
Also pinned the cylc-sphinx-extensions version to the upcoming release to ensure we don't get conflicts with the old sphinx config.

datamel (approved):

PR #153, opened 2020-08-21
glossary: fix reference

oliver-sanders:
Capitalisation was causing this reference to fail.


oliver-sanders:
Fix a couple of issues hanging over from the Sphinx upgrade.

PR #176, opened 2020-12-03
Config changes

wxtim:
Change a small number of links referring to config items which have been moved/renamed.

Partially closes_: https://github.com/cylc/cylc-flow/issues/3696

### DIFF
--- a/cylc/doc/etc/tutorial/runtime-introduction/.validate
+++ b/cylc/doc/etc/tutorial/runtime-introduction/.validate
@@ -21,7 +21,6 @@ set -eux
 TEMPORARYDIRECTORY=$(mktemp -dp "$HOME"/cylc-run/)
 cp -r ./* "$TEMPORARYDIRECTORY"
 cd "$TEMPORARYDIRECTORY"
-sed -i '1s;^;[scheduler]\n    abort if any task fails = True\n;' flow.cylc
 cylc validate --strict . --icp=2000
-cylc run --no-detach
+cylc run --no-detach --abort-if-any-task-fails
 rm -r "$TEMPORARYDIRECTORY"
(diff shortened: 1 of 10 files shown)


cylc/doc/etc/tutorial/runtime-introduction/.validate
```diff
@@ -21,7 +21,6 @@ set -eux
 TEMPORARYDIRECTORY=$(mktemp -dp "$HOME"/cylc-run/)
 cp -r ./* "$TEMPORARYDIRECTORY"
 cd "$TEMPORARYDIRECTORY"
-sed -i '1s;^;[scheduler]\n    abort if any task fails = True\n;' flow.cylc
```
MetRonnie, line 24: Yuk 😆 A lot nicer now

src/user-guide/running-suites.rst
```diff
@@ -644,7 +644,6 @@ These can be configured using the settings:
 
 - :cylc:conf:`[mail]to` (list of email addresses)
 - :cylc:conf:`[mail]from`
-- :cylc:conf:`[mail]smtp`
```
MetRonnie, line 647: The SMTP server can still be set, but only in global config. Maybe it's worth linking to that?

wxtim: I don't think so - I think that we'd rather most users don't tinker with the global config.

MetRonnie (approved):

src/user-guide/running-suites.rst
MetRonnie: Ok fair enough

cylc/doc/etc/tutorial/runtime-introduction/.validate
wxtim: It took me a while and some panicky message to Oliver to sort this. It's an important lesson with Sed though: Separators can be chars not="/", and it's often convenient if they are something else

kinow (approved):

PR #180, opened 2020-12-17
batch sys -> job runner

MetRonnie:
Sibling to https://github.com/cylc/cylc-flow/pull/3992

PR #182, opened 2020-12-17
Remove out-of-date quirk doc.

hjoliver:
This removes documentation of the "quirk" fixed by cylc/cylc-flow#3452

One review will do.


hjoliver:
(doc build fail will be fixed by #180 )

kinow (approved):
Ah! Good catch! +1

PR #186, opened 2020-12-30
Auto update copyright year

MetRonnie:
If we get this in before the new year, then we won't have to make any changes later.

### DIFF
--- a/.github/workflows/update_copyright.yml
+++ b/.github/workflows/update_copyright.yml
@@ -1,0 +1,68 @@
+# THIS FILE IS PART OF THE CYLC SUITE ENGINE.
+# Copyright (C) NIWA & British Crown (Met Office) & Contributors.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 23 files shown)


cylc/doc/etc/tutorial/cylc-forecasting-suite/bin/consolidate-observations
```diff
@@ -1,9 +1,21 @@
 #!/usr/bin/env python3
 # -*- coding: utf-8 -*-
 # -----------------------------------------------------------------------------
-# Copyright (C) 2012-2019 British Crown (Met Office) & Contributors - GNU V3+.
-# This is illustrative code developed for tutorial purposes, it is not
-# intended for scientific use and is not guarantied to be accurate or correct.
```
oliver-sanders, line 6: I think this was put here for a reason, best to keep.

.github/workflows/update_copyright.yml
```diff
@@ -0,0 +1,68 @@
+# THIS FILE IS PART OF THE CYLC SUITE ENGINE.
```
oliver-sanders, line 1: Incorrect, but appears in some other cylc-doc files too (probably my fault), can strip later.

cylc/doc/etc/tutorial/cylc-forecasting-suite/bin/consolidate-observations
MetRonnie: Addressed in https://github.com/cylc/cylc-doc/pull/186/commits/77d3efb90e0ba2df34c1f0c058dfe73b6643073c

oliver-sanders (approved):

PR #193, opened 2021-01-22
Add "nightly.latest" link

MetRonnie:
Had this idea after always having to update the URL of my ever-open nightly cylc-doc browser tab.

It's like how https://cylc.github.io/cylc-doc/current links to the latest version of the docs, but for the nightly builds, so that https://cylc.github.io/cylc-doc/nightly.latest links to the latest successful nightly build.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -103,15 +103,15 @@ jobs:
 
       - name: build docs
         run: |
-          build_name="nightly.$(isodatetime --utc -f %Y-%m-%d)"
           make -C docs \
             html \
             slides \
             linkcheck \
             SPHINXOPTS='-Wn' \
             SETCURRENT=false \
-            CYLC_VERSION="$build_name"
-          git -C gh-pages add "$build_name" 'versions.json'
+            BUILDDIR='doc/nightly' \
+            CYLC_VERSION="nightly.$(isodatetime --utc -f %Y-%m-%d)"
+          git -C gh-pages add 'nightly' 'versions.json'
 
       - name: push changes
         working-directory: gh-pages


oliver-sanders:
I think this will make the root `index.html` auto-redirect to the nightly build rather than the "current" one?

oliver-sanders:
Could just remove the date from the "nightly" dirname, I think this would do the trick:

```diff
diff --git a/.github/workflows/nightly.yml b/.github/workflows/nightly.yml
index 2f00a52..9e95ec9 100644
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -116,15 +116,14 @@ jobs:
 
       - name: build docs
         run: |
-          build_name="nightly.$(isodatetime --utc -f %Y-%m-%d)"
           make -C docs \
             html \
             slides \
             linkcheck \
             SPHINXOPTS='-Wn' \
             SETCURRENT=false \
-            CYLC_VERSION="$build_name"
-          git -C gh-pages add "$build_name" 'versions.json'
+            CYLC_VERSION="nightly.$(isodatetime --utc -f %Y-%m-%d)"
+          git -C gh-pages add 'nightly' 'versions.json'
 
       - name: push changes
         working-directory: gh-pages
```

MetRonnie:
Unfortunately that doesn't work. I've pushed up another commit which fixes the problem

oliver-sanders:
Ah, tis because the build dir is set from the `CYLC_VERSION` 🤦, can override using a make opt.

```
BUILDDIR=doc/nightly
```

MetRonnie:
Ah that's simpler. Force-pushed

oliver-sanders (approved):

oliver-sanders:
Thanks, I have no idea why I put the date in the URL, that was daft.

PR #199, opened 2021-02-12
Implement "stable" & "latest" doc builds

MetRonnie:
Add a "latest" build configuration. This will mean https://cylc.github.io/cylc-doc/latest/index.html can link to the latest pre-release. "current" is replaced by "stable", so both https://cylc.github.io/cylc-doc/stable/index.html and https://cylc.github.io/cylc-doc/index.html will link to the stable release.


.github/workflows/deploy.yml
```diff
@@ -72,7 +76,7 @@ jobs:
           path: gh-pages
 
       - name: sync static files
-        if: ${{ github.event.inputs.set_current }}
+        if: ${{ github.event.inputs.set_stable }}
         run: |
```
MetRonnie, line 80: I don't understand why this only happens for `.set_current == true`. Does this need to happen for `.set_latest` too?

oliver-sanders: I think we used to have other stuff in the `doc` folder including the "suite design guide" and other things that were version-pinned against Cylc which we wouldn't necessarily want to update for a pre-release.

Now the only thing we have there is the `joss-paper` so I guess we could update with every run, just need to make sure that doesn't mess up the version/format links.

oliver-sanders (commented):
Looks good to me.

Should add `latest` and `stable` to the version globby thinggy like you did for `nightly` so that they appear in the version selector.

Should also make a PR to the gh-pages branch to remove the current and add the stable and latest symlinks.

MetRonnie:
I also needed to modify the undeploy workflow https://github.com/cylc/cylc-doc/pull/199/commits/e3998c5d4fd288f6e9ef81347646efadb4a197c1#diff-691c7482d20538c7fc43a01c6722810dad9f756b7960755c53df313aa34c433c

bin/create-html-redirect
```diff
@@ -0,0 +1,35 @@
...
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
```
oliver-sanders, line 16: Ooh, better get this brexit ready:

```suggestion

set -eu
```

oliver-sanders (approved):

MetRonnie:
Todo:
- [ ] Make a PR to the gh-pages branch to remove the current and add the stable and latest symlinks.
- [ ] Update links in rose, e.g. here https://github.com/metomi/rose/blob/master/sphinx/hyperlinks.rst

PR #202, opened 2021-02-18
run/restart to play

datamel:
Checks on cylc doc were failing due to cylc run still being used in the tutorial suite. This is just a quick search/replace pr in an attempt to fix.


datamel:
@MetRonnie, this is not going to be a thorough documentation update for cylc play, just a quick one to try and get the build working, although I suspect it may still innately flakey.

MetRonnie (changes requested):
Actually, can I ask you to leave the rst files as they were? They might conflict with my `cylc play|pause` PR when it comes along, and it will be easier to rewrite sections if the original run vs restart context is still there

MetRonnie (approved):

MetRonnie:
Sorry for undoing most of the changes!

oliver-sanders (approved):

PR #203, opened 2021-02-23
actions: enable changing the branch of documented projects in testing

oliver-sanders:
Make it easier to ensure that changes to documented projects don't cause docs build failures.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -46,7 +46,7 @@ jobs:
       - name: install libs to document
         env:
           cylc_ref: ${{ github.event.inputs.cylc_ref || 'master' }}
-          rose_ref: ${{ github.event.inputs.cylc_ref || 'master' }}
+          rose_ref: ${{ github.event.inputs.rose_ref || 'master' }}
           cylc_rose_ref: ${{ github.event.inputs.cylc_rose_ref || 'master' }}
         run: |
           # NOTE: Install with [all] so we can import plugins which may
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,6 +5,15 @@ on:
     branches: [master]
   pull_request:
   workflow_dispatch:
+      rose_ref:
+        description: The Rose branch to test against
+        required: false
+      cylc_ref:
+        description: The Cylc branch to test against
+        required: false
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -5,6 +5,15 @@ on:
...
   workflow_dispatch:
+      rose_ref:
+        description: The Rose branch to test against
+        required: false
+      cylc_ref:
+        description: The Cylc branch to test against
+        required: false
+      cylc_rose_ref:
+        description: The Cylc branch to test against
+        required: false
```
MetRonnie, line 16: You could use `default: master` here instead of `${{ github.event.inputs.*_ref || 'master' }}` below

.github/workflows/test.yml
```diff
@@ -35,13 +44,17 @@ jobs:
           pip install .[all]
 
       - name: install libs to document
+        env:
+          cylc_ref: ${{ github.event.inputs.cylc_ref || 'master' }}
+          rose_ref: ${{ github.event.inputs.cylc_ref || 'master' }}
```
MetRonnie, line 49:
```suggestion
          rose_ref: ${{ github.event.inputs.rose_ref || 'master' }}
```

MetRonnie (changes requested):
One typo and an optional suggestion

.github/workflows/test.yml
oliver-sanders: This is how I initially tried it, however, the `github.event` "context" is only available when the workflow is run through the `workflow_dispatch` and is not present otherwise meaning that you can't use inputs as a convenient way to store defaults for inputs which is very annoying.

Could provide a "global" `env` section to centralise things.

MetRonnie (approved):

MetRonnie:
Assuming the test failure is unrelated, but will leave to you to merge just in case

oliver-sanders:
That's a new one, having a go at re-ordering the `apt-get` installs.

oliver-sanders:
Error might be due to missing deps, having a crack at fixing, though with actions changing the version of ubuntu you get it might be hard to tell what works :(

MetRonnie:
Try adding `gsfonts` to the `apt install` step

PR #205, opened 2021-03-01
Fix build failure and move apt install step to its own action

MetRonnie:
Fix ``convert-im6.q16: unable to read font `helvetica'`` error on nightly & deploy workflows

### DIFF
--- a/.github/actions/install-dependencies/action.yml
+++ b/.github/actions/install-dependencies/action.yml
@@ -1,0 +1,29 @@
+name: Install dependencies
+description: apt-get dependencies plus pygraphviz
+
+runs:
+  using: composite
+  steps:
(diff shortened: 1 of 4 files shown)


oliver-sanders:
I'm guessing the last fix only worked on certain versions of Ubuntu?

oliver-sanders (approved):
Works on Ubuntu 20.

oliver-sanders:
If we hit any more problems can consider installing `imagemagick` from apt or building the stack in Conda.

MetRonnie:
> I'm guessing the last fix only worked on certain versions of Ubuntu?

I think it worked for any current version, but it was only implemented for the test workflow, not nightly or deploy

PR #216, opened 2021-03-16
Remove documentation of removed CYLC_SUITE_DEF_PATH(_ON_SUITE_HOST) v…

wxtim:
…ariables.

### DIFF
--- a/src/user-guide/writing-suites/configuration.rst
+++ b/src/user-guide/writing-suites/configuration.rst
@@ -103,7 +103,7 @@ e.g.:
 
 .. code-block:: cylc
 
-   # include the file inc/foo.cylc:
+   # include the file ~/cylc-run/workflow/inc/foo.cylc:
    %include inc/foo.cylc
 
 .. note::
--- a/src/user-guide/writing-suites/configuration.rst
+++ b/src/user-guide/writing-suites/configuration.rst
@@ -103,7 +103,7 @@ e.g.:
 
 .. code-block:: cylc
 
-   # include the file $CYLC_SUITE_DEF_PATH/inc/foo.cylc:
+   # include the file inc/foo.cylc:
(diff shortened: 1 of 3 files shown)


src/user-guide/writing-suites/configuration.rst
```diff
@@ -103,7 +103,7 @@ e.g.:
 
 .. code-block:: cylc
 
-   # include the file $CYLC_SUITE_DEF_PATH/inc/foo.cylc:
+   # include the file inc/foo.cylc:
```
datamel, line 106: Should this be 
```suggestion
   # include the file ~/cylc-run/workflow/inc/foo.cylc:
```
The preceding line says: `Include-file paths can be specified portably relative to the suite configuration directory, e.g.:`  I wonder if it would be worth including the full path of the file so the relative path is more clear?

src/user-guide/writing-suites/runtime.rst
```diff
@@ -376,8 +372,7 @@ run time to the home directory of task owner on the task host.
 How Tasks Get Access To The Suite Directory
 -------------------------------------------
 
-Tasks can use ``$CYLC_SUITE_DEF_PATH`` to access suite files on
-the task host, and the suite bin directory is automatically added
+The suite bin directory is automatically added
```
datamel, line 375:
```suggestion
The workflow bin directory is automatically added
```
Although I guess the mass workflow/suite change has not been done so this is unimportant I guess!

datamel (approved):

src/user-guide/writing-suites/runtime.rst
wxtim: I'm going to have to keep correcting myself for years...

oliver-sanders (approved):

PR #212, opened 2021-03-15
quick tidy

oliver-sanders:
Built on #201 

* Merge the two different plugins sections I had created.
* Merge the three task-job sections under a single section.
* Add "this is beta software" warnings to the docs.

Structure a bit simpler now:

Screenshot from 2021-03-15 11-32-03

PR #219, opened 2021-03-19
Fix "make clean"

MetRonnie:
Added new paths to auto-generated content introduced by #212

https://github.com/cylc/cylc-doc/blob/8088b4f2c8a603744b7b0c444e79f69c0ac6885e/.gitignore#L24-L26

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,9 +19,8 @@ help:
 clean:
 	@$(SPHINXBUILD) -M clean "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
 	# remove auto-generated content
-	rm -rf src/user-guide/plugins/main-loop/built-in
-	rm -rf src/user-guide/batch-sys-handlers
-	rm -rf src/user-guide/job-runner-handlers
+	rm -rf src/plugins/main-loop/built-in
+	rm -rf src/user-guide/task-implementation/job-runner-handlers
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -19,6 +19,9 @@ help:
 clean:
 	@$(SPHINXBUILD) -M clean "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
 	# remove auto-generated content
+	rm -rf src/plugins/main-loop/built-in
+	rm -rf src/user-guide/task-implementation/job-runner-handlers
+	# remove auto-generated content (old)
 	rm -rf src/user-guide/plugins/main-loop/built-in
```
oliver-sanders, line 25:
```suggestion
```

Makefile
```diff
@@ -19,6 +19,9 @@ help:
...
 	rm -rf src/user-guide/batch-sys-handlers
 	rm -rf src/user-guide/job-runner-handlers
```
oliver-sanders, line 27:
```suggestion
```

MetRonnie: The idea was to remove old files that might be kicking around due to checking out between different branches, is that not needed?

oliver-sanders: Can do with `git clean` rather than leaving old paths in a makefile.

oliver-sanders (approved):

PR #224, opened 2021-03-23
flow terminology

oliver-sanders:
We've been playing with the word "flow" for long enough, time to get something solid down on paper so we can get on with the renaming and documentation.

Here are the terms in question (see also - https://github.com/cylc/cylc-flow/issues/3817#issuecomment-717206611):

* Workflow (run dir).
* Source Workflow (source dir or arbitrary location).
* Flow / Re-Flow.
* Workflow ID (i.e. registration).

> **Note:** Some glossary definitions missing, will fill these in once the "flow" terminology is agreed.

### DIFF
--- a/src/glossary.rst
+++ b/src/glossary.rst
@@ -5,10 +5,35 @@ Glossary
    :sorted:
 
    workflow
-   Cylc workflow
-      A Cylc workflow is defined by a :cylc:conf:`flow.cylc` file
-      containing a dependency :term:`graph<graph>` and :term:`task`
-      definitions.
+   cylc workflow
+      A Cylc workflow is a collection of :term:`tasks <task>` to carry out and
+      :term:`dependencies <dependency>` that govern the order in which they
+      run. This is represented in Cylc format in a :cylc:conf:`flow.cylc` file.
+
+      For example here is a Cylc workflow representing the brewing process:
+
+      .. code-block:: cylc
+         :caption: flow.cylc
+
+         [scheduling]
+             cycling mode = integer
+             initial cycle point = 1
+             [[graph]]
+                 # repeat this for each batch
+                 P1 = """
+                     # the stages of brewing in the order they must occur in
+                     malt => mash => sparge => boil => chill => ferment => rack
+
+                     # must finish the sparge of one batch before
+                     # starting the next one
+                     # sparge[-P1] => mash
+                 """
+
+      .. admonition:: Cylc 7
+         :class: tip
+
+         In Cylc version 7 and earlier "workflows" were referred to as
+         "suites".
 
    graph
       The graph of a :term:`workflow<Cylc workflow>` refers to the
@@ -422,26 +447,6 @@ Glossary
 
       * :ref:`ImplicitTasks`
 
-   run directory
-      The run directory contains all of the configuration for a workflow, e.g.
-      the :cylc:conf:`flow.cylc` file.
-
-      It contains all the necessary files to run the workflow and typically
-      resides in the :term:`cylc-run directory`:
-
-      ``~/cylc-run/<workflow-name>``
-
-      The run directory can be accessed by a running workflow using
-      the environment variable ``CYLC_WORKFLOW_RUN_DIR``.
-
-      See also:
-
-      * :term:`source directory`
-      * :term:`work directory`
-      * :term:`share directory`
-      * :term:`job log directory`
-      * :term:`cylc-run directory`
-
    work directory
       When Cylc executes a :term:`job` it does so inside the
       :term:`job's <job>` working directory. This directory is created by Cylc
@@ -524,13 +529,14 @@ Glossary
       workflows.
 
    contact file
-      The contact file records information about a running workflow such as the host it
-      is running on, the TCP port(s) it is listening on and the process ID.
-      The file is called ``contact`` and lives inside the workflow's
+      The contact file records information about a running workflow such as the
+      host it is running on, the TCP port(s) it is listening on and the process
+      ID. The file is called ``contact`` and lives inside the workflow's
       :term:`service directory`.
 
-      The contact file only exists when the workflow is running, if you delete the
-      contact file, the workflow will (after a delay) notice this and shut down.
+      The contact file only exists when the workflow is running, if you delete
+      the contact file, the workflow will (after a delay) notice this and shut
+      down.
 
       .. warning::
 
@@ -716,14 +722,67 @@ Glossary
       * :term:`run directory`
 
    source directory
-      Any directory where workflows are written and stored in preparation for
-      installation with ``cylc install`` or reinstallation with
-      ``cylc reinstall``.
+      Any directory where :term:`workflows <workflow>` are written and stored
+      in preparation for installation with ``cylc install`` or reinstallation
+      with ``cylc reinstall``.
+
+      .. tip::
+
+         You can configure the default locations where the ``cylc install``
+         will look for source directories using the
+         :cylc:conf:`global.cylc[install]source dirs` configuration.
 
       See also:
 
       * :term:`run directory`
       * :ref:`Installing-workflows`
+
+   run directory
+      This is a directory containing the configuration that Cylc uses to run
+      the :term:`workflow`.
+
+      Typically this is installed from the :term:`source directory` using
+      ``cylc install``.
+
+      The run directory can be accessed by a running workflow using
+      the environment variable ``CYLC_WORKFLOW_RUN_DIR``.
+
+      See also:
+
+      * :term:`source directory`
+      * :term:`work directory`
+      * :term:`share directory`
+      * :term:`job log directory`
+      * :term:`cylc-run directory`
+
+   play
+      We run a :term:`workflow` using the ``cylc play`` command.
+
+      This starts a :term:`scheduler` which is the program that controls the
+      flow and is what we refer to as "running".
+
+      You can :term:`play`, :term:`pause` and :term:`stop` a :term:`flow`,
+      Cylc will always carry on where it left off.
+
+      See also:
+
+      * :term:`pause`
+      * :term:`stop`
+
+   pause
+      When a :term:`workflow` is "paused" the :term:`scheduler` is still
+      running, however, will not submit any new jobs.
+
+      This can be useful if you want to hold the workflow back whilst you make
+      a change.
+
+      Pause a workflow using ``cylc pause`` and unpause it using ``cylc play``.
+
+      See also:
+
+      * :term:`play`
+      * :term:`stop`
+      * :term:`hold`
 
    stop
    shutdown
@@ -737,6 +796,8 @@ Glossary
 
       See also:
 
+      * :term:`play`
+      * :term:`pause`
       * :ref:`Stopping Suites`
       * :term:`start`
       * :term:`restart`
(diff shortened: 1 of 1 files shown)


src/glossary.rst
```diff
@@ -967,3 +967,115 @@ Glossary
       See also:
 
       * :ref:`Graph Branching`
+
+source workflow
+   A source workflow is a :term:`cylc workflow` that has not yet been
+   :term:`installed` into the :term:`run directory`.
```
MetRonnie, line 973: Only thing is, the source workflow still exists separately after it has been installed.

src/glossary.rst
```diff
@@ -967,3 +967,115 @@ Glossary
...
+
+      .. ..note:: For convenience you can omit the ``/run1`` on the command line.
+
+flow
+   A flow is a single logical run of a :term:`workflow` that is done by
+   a :term:`scheduler`.
+
+   A flow can be :term:`played <play>` and :term:`paused <pause>`,
+   :term:`started <start>` and :term:`stopped <stop>`, it is still the same
+   flow.
```
MetRonnie, line 1023:
```suggestion
   :term:`started <start>` and :term:`stopped <stop>`, and will remain
   the same flow.
```

src/glossary.rst
```diff
@@ -967,3 +967,115 @@ Glossary
...
+      For example, if your workflow is in ``~/cylc-run/foo/bar/run1`` then its ID
+      is ``foo/bar/run1``.
+
+      .. ..note:: For convenience you can omit the ``/run1`` on the command line.
+
+flow
+   A flow is a single logical run of a :term:`workflow` that is done by
+   a :term:`scheduler`.
+
+   A flow can be :term:`played <play>` and :term:`paused <pause>`,
```
MetRonnie, line 1021: I don't think there is a defintion for "play" at the moment, I suppose it's "causing the scheduler to run, either by starting it, restarting it or resuming it"

oliver-sanders: I think there are a few definitions missing ATM, waiting to get the "flow" stuff sorted before filling in the blanks.

src/glossary.rst
oliver-sanders: Yep, on second read this isn't very good, attempt two.

src/glossary.rst
```diff
@@ -967,3 +967,117 @@ Glossary
...
+
+         [scheduling]
+             cycling mode = integer
+             initial cycle point = 1
+             [[graph]]
+                 P1 = """
+                     b[-P1] | c[-P1] => a
+                     a:succeed => b
+                     a:fail => c
+                 """
```
hjoliver, line 1054: We'll need to start adding `allow implicit tasks = True` to all our example workflows so that they can run out of the box.

src/glossary.rst
```diff
@@ -967,3 +967,117 @@ Glossary
...
+
+      Let's say the workflow has run to cycle 8, but we have just noticed that
+      something went wrong with the task ``a`` in cycle 5.
+
+      To rectify this we could create a new flow (a reflow) starting at ``a.5``::
+
+         # get the ID of the original flow
+         cylc TBC <id> a.5  # ???
```
hjoliver, line 1062: At the moment, have to get it from the log:
>  (a.10) spawned b.10 flow(z)

src/glossary.rst
```diff
@@ -967,3 +967,117 @@ Glossary
...
+
+         # pause the original flow
+         cylc pause <id> --flow <flow>
```
hjoliver, line 1065: We can't do this yet.  We need to decide on scheduler pause vs task hold implementation. Probably the latter, i.e. we'd end up with held tasks from the targeted flow, in which case this example should use the `cylc hold` command.

hjoliver: We can already `cylc stop` a particular flow, in which case current active tasks in the flow will not spawn downstream children.

src/glossary.rst
```diff
@@ -967,3 +967,117 @@ Glossary
...
+
+         # create a new flow starting at a.5
+         cylc trigger a.5 --reflow
+
+      Where ``<id>`` is the name of the :term:`workflow` and ``<flow>`` is the
+      identifier of the :term:`flow`.
+
+      This new flow will run over the top of the old one.
```
hjoliver, line 1073: If the old flow was *stopped*, the new flow will carry on past the point where the old flow stopped.

If the old flow was *held*(/(*paused*?), the new flow will run into the held tasks of the new flow, and the flow IDs will merge there.

src/glossary.rst
hjoliver: Also, holding tasks is currently problematic: cylc/cylc-flow#3743

src/glossary.rst
oliver-sanders: Ach, dammit, that makes examples so convoluted.

src/glossary.rst
oliver-sanders: If `cylc hold '*'` targets all tasks in the n=1 window then that would do the trick.

(off topic for this PR, n=1 matching + a new list in the task_pool caching hold requests for proxies that aren't in the pool yet backed up by a new DB table should do the trick)

src/glossary.rst
oliver-sanders: Explicitly stopping is probably un-necessary, however, a bit cleaner?

If the old and new flows take different paths a `cylc release '*'` could potentially result in the old flow continuing?

hjoliver: > Explicitly stopping is probably un-necessary, however, a bit cleaner?

Yes, but easier to achieve, hence we have it already :grin: 

> If the old and new flows take different paths a cylc release '*' could potentially result in the old flow continuing?

Yes (unless you "stop" it)

src/glossary.rst
```diff
@@ -967,3 +967,117 @@ Glossary
       See also:
 
       * :ref:`Graph Branching`
+
+source workflow
```
wxtim, line 971: workflow source?

oliver-sanders: The source workflow is in the workflow source.

src/glossary.rst
```diff
@@ -967,3 +967,117 @@ Glossary
...
+   A source workflow is a :term:`cylc workflow` that is not located in the
+   :term:`cylc run directory`.
+
+   In order to be :term:`run` source workflows must be
+   :term:`installed <install>` into the :term:`cylc run directory`.
+
+workflow
+cylc workflow
+   A Cylc workflow is a collection of things to run (:term:`tasks`) along with
```
wxtim, line 980: "things" → "tasks"

src/glossary.rst
```diff
@@ -967,3 +967,117 @@ Glossary
...
+   the constraints that govern the order in which they run
+   (:term:`dependencies`) represented in Cylc format in a
+   :cylc:conf:`flow.cylc` file.
+
+   For example here is a Cylc workflow representing the brewing process:
```
wxtim, line 985: 🤣 How very you....

src/glossary.rst
wxtim: Would these emoji's potentially make our conceit RE: The control terminology more obvious? ⏯️ ⏹️

src/glossary.rst
wxtim: Perhaps we should jot some abstract examples down on cylc-admin:
Off the top of my head:
- Cooking
- Gardening 
- DIY
- Start of Driving lesson (chair, seatbelt, mirrors, hands at 10 and 2, clutch (if manual), choke(if old car), start, mirror, signal, maneuver, possible branch )
-

src/glossary.rst
wxtim: or define the tasks in the runtime section?

src/glossary.rst
wxtim: If you can't play it then surely it's not a workflow?

MetRonnie: Middle ground: workflow source workflow?

hjoliver: Programming analogies might help...

IMO "source code" refers generically to the medium. i.e., that bunch of text is "source code".  It doesn't mean, that is the source for this particular computer program. For that, you would say "program source", not "source program" which sounds like "a program that is the source of something else", not "the source code of the program".

So on that basis, I think we should go with:

*The **workflow source directory" contains the workflow config file that gets installed to (and is therefore the source of) this workflow run directory.*

We don't necessarily have to confuse people by refering to the workflow config file itself as a "workflow source" or "source workflow".

src/glossary.rst
hjoliver: > a disturbingly small proportion of the population are familiar with the brewing process

But a gratifying large proportion are familiar with the products thereof :beers:

hjoliver:
Where I've landed:

- a **workflow** is an abstract graph of dependent tasks that you want to run
  - it is defined in a workflow config file
  - it exists even when the workflow is not running
- a **scheduler** is a program that runs a workflow
  - when a scheduler is running a workflow, users interact with the scheduler, not with "the workflow" 
    - however it is reasonable to refer to e.g. "stopping the workflow" rather than "stopping the scheduler" because a scheduler only runs one workflow
- a **flow** is a single self-consistent front of activity within a running workflow
   - there can be multiple flows at once in a running workflow  
- **reflow** describes the process of triggering a new flow within an already-running workflow 
   - the "re" bit is because reflows are almost always in the past graph, where the original flow has already passed through

hjoliver:
In light of the above, to be as clear and consistent as possible with our terminology we should really rename `flow.cylc` to `workflow.cylc`.

It's not too late to do that as almost no one outside the dev team is using Cylc 8 yet. But if we do it, we should do it very soon!

wxtim:
> In light of the above, to be as clear and consistent as possible with our terminology we should really rename `flow.cylc` to `workflow.cylc`.
> 
> It's not too late to do that as almost no one outside the dev team is using Cylc 8 yet. But if we do it, we should do it very soon!
> 
> 

It's a lot of extra chars, but tab completion; w is such an unusual letter it may actually be easier.

hjoliver:
Or we could keep `flow.cylc` but say "flow" is just convenient shorthand for "workflow" in this context?

src/glossary.rst
oliver-sanders: Even worse?

src/glossary.rst
oliver-sanders: (resolving this as the code is a comment that is not rendered, we can fix this once we have reflow sorted)

src/glossary.rst
```diff
@@ -4,10 +4,30 @@ Glossary
...
 
-   suite
-   Cylc suite
-      A Cylc suite is a directory containing a :cylc:conf:`flow.cylc` file
-      which contains :term:`graphing<graph>` representing a workflow.
+   workflow
+   cylc workflow
+      A Cylc workflow is a collection of :term:`tasks` to carry out and
+      :term:`dependencies` that govern the order in which they run. This is
+      represented in Cylc format in a :cylc:conf:`flow.cylc` file.
```
MetRonnie, line 11: How about a note that says workflows used to be called suites?

src/glossary.rst
```diff
@@ -707,15 +713,58 @@ Glossary
       * :term:`run directory`
 
    source directory
-      Any directory where workflows are written and stored in preparation for
-      installation with ``cylc install`` or reinstallation with
-      ``cylc reinstall``.
+      Any directory where :term:`workflows <workflow>` are written and stored
+      in preparation for installation with ``cylc install`` or reinstallation
+      with ``cylc reinstall``.
```
MetRonnie, line 718: Might be worth mentioning the search path can be configured in `global.cylc[install]source dirs`

src/glossary.rst
```diff
@@ -707,15 +713,58 @@ Glossary
...
+
+      You can :term:`play`, :term:`pause` and :term:`stop` a :term:`flow`,
+      Cylc will always carry on where it left off.
+
+      See also:
+
+      * :term:`pause`
+      * :term:`stop`
+
+   pause
```
MetRonnie, line 754: just realised, there's already a definition for pause on master, is this branch a bit behind?

MetRonnie (changes requested):

oliver-sanders:
> We should try to get this PR merged soon

Now failing, how about merging it into your PR?

MetRonnie (approved):

hjoliver (approved):
:+1:

PR #234, opened 2021-04-01
proposal: meta package

oliver-sanders:
sibling: https://github.com/cylc/cylc-flow/pull/4158

* Documents the meta-package.
* Adds some missing Cylc Rose docs.

### DIFF
--- a/src/installation.rst
+++ b/src/installation.rst
@@ -30,46 +30,68 @@ present Cylc only tested on Linux.
 Quick Installation
 ------------------
 
-Via Conda (recommended):
+Conda (recommended)
+^^^^^^^^^^^^^^^^^^^
+
+Base installation (just the :term:`scheduler`):
 
 .. code-block:: console
 
-   # install cylc with the browser-GUI
    $ conda install cylc
 
-Via pip:
-
-.. note::
-
-   We recommend using a virtual environment.
-
-.. note::
-
-   Requires Python 3.7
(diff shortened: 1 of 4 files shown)


src/installation.rst
```diff
@@ -30,47 +30,69 @@ present Cylc only tested on Linux.
...
+
+   You must ensure :ref:`non-Python dependencies` have been installed via
+   other means.
 
-   Requires Python 3.7
+.. tip::
+   .. _venv: https://docs.python.org/3/tutorial/venv.html
+   .. _pipenv: https://pipenv.pypa.io/en/latest/
+
+   We recommend using a virtual Python environement such as `venv`_ or
```
hjoliver, line 61:
```suggestion
   We recommend using a virtual Python environment such as `venv`_ or
```

src/installation.rst
```diff
@@ -30,47 +30,69 @@ present Cylc only tested on Linux.
...
+   $ conda install cylc cylc.uiserver cylc.rose
 
-.. note::
+Pip
+^^^
+
+.. warning::
+
+   You must ensure :ref:`non-Python dependencies` have been installed via
+   other means.
```
hjoliver, line 55: Mention that a built cylc-ui is now included in cylc-uiserver?

hjoliver: (OK I see you're not including cylc-ui in the list of non-Python deps, further down).

Issue #185, opened 2020-12-23
autodocument CLI


oliver-sanders:
depends on: https://github.com/cylc/cylc-flow/issues/2972
related to: https://github.com/cylc/cylc-doc/issues/11

Autodocument the CLI.

Requires the CLI to be implemented in something that is autodocumentable by Sphinx (e.g. `agparse`, `click`) or the implementation of a crude autodocumenter (like we used to have).

PR #237, opened 2021-04-13
Restore command help reference section

hjoliver:
As far as I recall we didn't decided to ditch the CLI command help reference section of the User Guide??

Maybe we just disabled it when the generator script broke after removing the old command "categories" in the cylc-flow CLI. 

If we want to restore it, it's an easy fix.   Any strong opinions?

### DIFF
--- a/bin/autodoc-cli
+++ b/bin/autodoc-cli
@@ -15,11 +15,10 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-# Create appendices/command-ref.rst for inclusion in HTML doc.
+# Create command help transcripts for inclusion in HTML doc.
 
 # All paths relative to 'doc/src/custom/' directory:
-COMMAND_REF_FILE="$(dirname "$0")/../src/appendices/command-ref.rst"
-CYLC="cylc"
+COMMAND_REF_FILE="$(dirname "$0")/../src/reference/command-help.rst"
 
 cat > "$COMMAND_REF_FILE" <<END
 .. _CommandReference:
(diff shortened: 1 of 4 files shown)


hjoliver:
Building the docs on this branch revealed https://github.com/cylc/cylc-flow/issues/4171

```
$ cylc play --help --colour=never
...
```
shot

hjoliver:
shot

bin/autodoc-cli
```diff
@@ -15,66 +15,30 @@
...
-
-cat >> "$COMMAND_REF_FILE" <<END
-
-Commands
---------
-
 END
 
-for COM in $($CYLC commands); do
+for COM in $(cylc help all | awk '{print $1}' | sort); do
```
hjoliver, line 31: CI build failed, I think because this line doesn't work if the terminal width is too low ?? (newlines get inserted into the help text)

oliver-sanders:
This was kinda purposefully removed pending conversion to `argparse` which would enable us to autodocument using standard tooling - https://github.com/cylc/cylc-doc/issues/185

Of course we can keep the old CLI help for now though but be aware that all of the references will change when we change implementation.

hjoliver:
Oh yeah. From that Issue, I guess this PR restores the option of final resort:
>  [last resort] the implementation of a crude autodocumenter (like we used to have).

Looking at the command ref section (or the image above) it probably is quite useful to have.  All the command names are visible at once, and you can click on them to instantly see what they do (to state the obvious). 

If the aim is to retain the command reference but generate it by more sophisticated means, then I think we should keep the old way until we have that. Users don't care about the implementation.

oliver-sanders:
Fair enough, we can put it back in for now with the caveat that anyone who links into this page has automatically volunteered to fix the references later 😈.

(the main reason I pulled it out before putting the new one in is because it took an age to run for each build).

bin/autodoc-cli
oliver-sanders: Would have to be narrower than the longest command name to trip awk up?!

hjoliver:
It is slow. On my box 12 secs. (Out of an 18 sec build time).

hjoliver:
Closing, there's no great need for this.

PR #246, opened 2021-06-02
Fix overflowing diagrams

### LINKED ISSUE #209: overflowing diagrams
Some of the graphviz generated diagrams are overflowing:

overflow

* I think this may be a change in Sphinx? (I can't imagine it is a change in graphviz?).
* I think this effects graphs which have explicitly set the graphviz `size` attribute.

Find out what's going on. Can we remove the `size` attr, otherwise try to find sensible new values (make sure they work when you resize the page [at least as much as can be expected]).

MetRonnie:
Closes #209 

Use `ubuntu-18.04` in GitHub Actions instead of latest so that the graphviz version we get isn't buggy. See https://github.com/cylc/cylc-doc/issues/209#issuecomment-853248945

I tested this using the nightly workflow and it worked (however, because the workflow ran as scheduled overnight using the master branch (with `ubuntu-latest`), it won't show up).

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -44,7 +44,7 @@ on:
 
 jobs:
   deploy:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
     timeout-minutes: 10
     steps:
       - name: configure python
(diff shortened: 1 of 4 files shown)


oliver-sanders (approved):
Let's give it a twirl.

oliver-sanders:
Happy days https://cylc.github.io/cylc-doc/nightly/html/glossary.html#term-cycle

PR #249, opened 2021-06-07
Update docs for cylc hold

MetRonnie:
Twinned with cylc/cylc-flow#4238

Plus reword a couple of things to avoid confusion with cylc hold.

P.S. there isn't much mention of cylc hold in the docs, perhaps the tutorial could do with a cylc hold entry?

### DIFF
--- a/src/7-to-8/caveats.rst
+++ b/src/7-to-8/caveats.rst
@@ -35,7 +35,7 @@ CLI Change
 
       $ cylc stop '*'  # stop everything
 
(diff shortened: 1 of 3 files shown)


hjoliver:
> P.S. there isn't much mention of cylc hold in the docs, perhaps the tutorial could do with a cylc hold entry?

Yeah quite possibly. I think we can pick this up with a general review of the docs at some point soon.

hjoliver (approved):
:+1:

Issue #170, opened 2020-11-05
7-to-8 migration guide


hjoliver:
Check list of items that will need to be noted (ADD TO THIS LIST AS NEEDED)
- [x] `suite.rc` -> `flow.cylc`
- [x] restart from Cylc 7 DB not supported
- [x] new run semantics (`cylc play` etc.; and the `run1, run2, ...` sub-dirs
- [x] new scheduling section syntax
- [x] deprecated config items: recommend upgrading them
- [x] task state name translation (was 13 states, now 8)
- [x] named state checkpoints gone 
   - https://github.com/cylc/cylc-flow/issues/3891
   - recommend use of reflow-based warm start: https://github.com/cylc/cylc-flow/issues/3864#issuecomment-708699402
- [x] platforms concept explainer

Transferred from @oliver-sanders comment below:

Suggested topics:

* [x] cylc install (w/ note on `[install]source dirs` & run numbs/names "safe run semantics")
* [x] platforms
* [x] play / pause / stop (w/ note about hold/release)
* [x] reflow
* [x] remote installation (w/ note on `[scheduler]install` for people who used to use `rose suite-run`)
* [x] runahead limit changes (w/ recommendation to use queues where possible)
* [x] symlink dirs (for people who used the Rose `root dir` config to do this, note per-workflow conf is no longer possible) https://github.com/cylc/cylc-doc/pull/369
* [x] suicide triggers #241
* [x] suicide triggers update post possible SoD changes.
* [x] rose suite-run to cylc install migration (now remembers options)
* [x] rose stem (now provided by cylc install)
* [x] `allow implicit tasks` (note: doesn't affect Rose2019 users) (note: add link to migration page in warning message). https://github.com/cylc/cylc-doc/pull/304 https://github.com/cylc/cylc-flow/pull/4572

oliver-sanders:
Added a section for documenting Cylc 8 changes in https://github.com/cylc/cylc-doc/pull/241, just need to populate it!

Suggest the following format:

1. An admonition at the top of the page quickly conveying whether this is something the reader should delve into.
2. 1-2 line description of functionality if necessary.
3. Quick description of the change.
4. Example diff if required.

Example - https://github.com/cylc/cylc-doc/pull/241/files#diff-705a4752ecedd74e974b38897f45a8313a1042872eb7ead9056dd54080a37b61

These sections should be short, avoid documenting the actual feature here, just convey the change and link through to the main documentation.

[...]

PR #250, opened 2021-06-08
migration guide: template variables

oliver-sanders:
Related to https://github.com/cylc/cylc-doc/issues/170

Document the change to template vars.


hjoliver (approved):
:+1: builds fine, looks good.

PR #251, opened 2021-06-09
Fix small error in the text.

hjoliver:
One review will do.


kinow (approved):
Looks like at some point the for loop used integers, then it was updated to use booleans but the text was never touched. LGTM

PR #253, opened 2021-06-21
Remove obsoletion flow.cylc visualization section.

hjoliver:
Sibling of https://github.com/cylc/cylc-flow/pull/4273

Issue #168, opened 2020-10-28
HTTP 404 when changing the selected Cylc version on some pages


MetRonnie:
If you choose a different version of Cylc in the sidebar while browsing a page that does not exist on that version, you will get a 404 Not Found error. It would be good if it instead redirected you to the index or something. However I'm guessing this won't be feasible due to the limitations of hosting on GitHub Pages...

MetRonnie:
If we could somehow have a mapping of some of the most important pages e.g.
- [nightly/html/reference/config/workflow.html](https://cylc.github.io/cylc-doc/nightly/html/reference/config/workflow.html) <-> [stable/html/appendices/suiterc-config-ref.html](https://cylc.github.io/cylc-doc/stable/html/appendices/suiterc-config-ref.html)
- [nightly/html/reference/config/global.html](https://cylc.github.io/cylc-doc/nightly/html/reference/config/global.html) <-> [stable/html/appendices/site-user-config-ref.html](https://cylc.github.io/cylc-doc/stable/html/appendices/site-user-config-ref.html)

This could be made use of in the Makefile

PR #257, opened 2021-07-02
Redirect a selection of old pages to new ones

MetRonnie:
This partially addresses #168

Redirects the most important pages in the Cylc 7 and Cylc 8 early beta docs to their current counterparts, so when you switch versions from those to latest, you won't face a 404 anymore.

I think it's possible to do the opposite, but would require manually adding the current counterparts as html files that redirect to the old pages, in the Cylc 7 folders on the `gh-pages` branch

(Tests will fail until #253 merged)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,6 +39,8 @@ cleanall:
 	bin/version write > doc/versions.json
 	# Redirect doc/<version>/index.html -> doc/<version>/html/index.html
 	bin/create-html-redirect "html/index.html" "$(BUILDDIR)/index.html"
+	# Redirect old pages
+	bin/redirect-old-pages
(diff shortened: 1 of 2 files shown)


oliver-sanders (approved):
Looks good.

MetRonnie:
Test failure seems to be because it's running `redirect-old-pages` twice: for both `html` and `slides` in
```
make html slides linkcheck doctest SPHINXOPTS='-Wn'
```
I'm not sure how to make it run once

oliver-sanders:
Could convert this logic into JS and inject it into the `404` file, slightly ugly but it saves dumping redirect files about the place and makes redirecting files from removed directories a little easier.

https://github.com/cylc/cylc.github.io/blob/master/404.html

PR #283, opened 2021-08-05
release: 8.0b2

oliver-sanders:
* update installation instructions *not* to use the meta package
* https://github.com/cylc/cylc-admin/issues/130
* update installation instructions
* document instaling the cylc-rose package

### DIFF
--- a/src/installation.rst
+++ b/src/installation.rst
@@ -26,7 +26,7 @@ Cylc runs on Unix systems including Linux and Mac OS.
 .. list-table::
    :class: grid-table
 
-   * - .. rubric:: Via Conda (recommeneded):
+   * - .. rubric:: Via Conda (recommended):
      - .. rubric:: Via Pip (+npm):
 
    * - ::
--- a/src/installation.rst
+++ b/src/installation.rst
@@ -10,64 +10,75 @@ Installation
    It has a fully functional Python 3 workflow service and CLI that can run
    existing Cylc workflows.
 
-   **BUT:**
+   **But** it is not production-ready yet.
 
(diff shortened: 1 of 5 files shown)


src/installation.rst
```diff
@@ -10,64 +10,75 @@ Installation
...
 
+.. highlight:: sub
 
-Quick Installation
-------------------
+.. list-table::
+   :class: grid-table
 
-Via Conda (recommended):
+   * - .. rubric:: Via Conda (recommeneded):
```
MetRonnie, line 29:
```suggestion
   * - .. rubric:: Via Conda (recommended):
```

src/installation.rst
```diff
@@ -126,6 +137,8 @@ Cylc is split into a number of packages providing different functionality:
    Provides the scheduler "kernel" of Cylc along with the command-line.
 `Cylc UI Server`_
    Provides the "Cylc Hub" and the browser-based "Cylc GUI".
+:ref:`Cylc Rose`
+   Provides support for :ref:`Rose` suite configurations in Cylc workflows.
```
MetRonnie, line 141: (Slight niggle, probably not worth following up)

The link for Cylc Rose gets an underline whereas the others don't

image

oliver-sanders: Yep, that's because it is a `:ref:` rather than a hyperlink and the Sphinx RTD theme has decided that one should get an underline and not the other.

Nothing much we can do without hacking their CSS.

kinow (approved):

MetRonnie (approved):

PR #291, opened 2021-08-27
Only test tutorial workflow on PR when relevant files changed

MetRonnie:
Small change with no associated issue

### DIFF
--- a/.github/workflows/test_tutorial_workflow.yml
+++ b/.github/workflows/test_tutorial_workflow.yml
@@ -4,6 +4,12 @@ on:
   push:
     branches: [master]
   pull_request:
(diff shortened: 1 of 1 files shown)


oliver-sanders:
Not so sure this is a good idea as we will likely break the tutorial workflow with changes to cylc-flow unless we have scheduled testing and act when problems occur.

wxtim:
I'm afraid I agree with @oliver-sanders

MetRonnie:
That's why I left the tutorial tests on for pushes to master (which includes merged PRs), but I guess it still reduces visibility. Happy to live with the flakiness for now.

PR #294, opened 2021-09-02
move cylc.flow.processed

wxtim:
companion to https://github.com/cylc/cylc-flow/pull/4389


hjoliver (approved):
:+1: tested as working, one review will do.

PR #289, opened 2021-08-26
added a guide to writing a platform configuration.

wxtim:
Partially closes #254
Explains so example platform configurations.

Other bits
- Insert some description of using platforms and platform groups into user documentation. (Administrator docs in #289 )

- Also (totally off topic, and I have no idea how it ended up in this PR, additional info RE: Changes to graph syntax from https://github.com/cylc/cylc-flow/pull/4404

PR #299, opened 2021-09-22
Selection of hosts and platforms

wxtim:
Insert some description of using platforms and platform groups into user documentation. (Administrator docs in #289 )

Also (totally off topic, and I have no idea how it ended up in this PR, additional info RE: Changes to graph syntax from https://github.com/cylc/cylc-flow/pull/4404


wxtim:
folded into #289

Issue #271, opened 2021-07-16
Document Configurable Authorization 


datamel:
Documentation update needed in light of https://github.com/cylc/cylc-uiserver/pull/204

oliver-sanders:
Note if we document this in the traitlet this issue will be closed by #222

datamel:
Closed by https://github.com/cylc/cylc-uiserver/pull/300

PR #300, opened 2021-09-28
Authorization documentation

datamel:
Accompanies https://github.com/cylc/cylc-uiserver/pull/204.

Partially addresses Document Configurable Authorization #271.

Documents authorization and adds a new section, `Sharing Workflows` to the user-guide.

### DIFF
--- a/src/user-guide/sharing-workflows/sharing-workflows.rst
+++ b/src/user-guide/sharing-workflows/sharing-workflows.rst
@@ -185,11 +185,9 @@ Sites set both limits and defaults for users.
 - `default` sets a default access level, which is activated if the user does
   not appear in the user-authorization configuration.
 
-
 Missing Configurations in Site Authorization
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-
-Site configuration 
+For site configuration:
 
 * if a limit is not set but a default is, then the limit is the default.
 * if a default is not set but a limit is, then the default is no access.
--- a/src/user-guide/sharing-workflows/sharing-workflows.rst
+++ b/src/user-guide/sharing-workflows/sharing-workflows.rst
@@ -1,0 +1,284 @@
+.. _Authorization:
+
+Authorization
+=============
+
+Cylc supports multi user access. You can grant control of your
+workflows to other users, by adding or removing privileges on a Cylc operation
+basis.This requires site level configuration, with individual users being able
+to set authorization settings, within the bounds set by the site.
+
+Sites can set default access rights for users, for more information, see
+:ref:`site_configuration`.
+
+Note, if this feature is not configured, the default is no multi user access
+and access will be limited to the workflow owner.
+
+If you grant access to another user, this access will apply to all workflows.
+
+Granting Access
+---------------
+
+There are three methods of identifying a user to grant access to:
+
+- ``*`` character, to indicate any authenticated user,
+- using the ``group:`` prefix to indicate a system group. E.g.
+  ``group:groupname``, will assign permissions to all members of the system
+  group named `groupname`. For more information, see :ref:`group_support`.
+- ``username`` to indicate a specific user.
+
+All permissions are additive, if user appears elsewhere in configuration, for
+example as a member of a system group, the permission level is taken as the
+greatest possible.
+
+However, any negations are applied with priority.
+
+Note, default in site config does not contribute to permissions, this is used
+in the case where a user does not appear in the `c.CylcUIServer.user-authorization`.
+
+.. note::
+
+   Spellings of UIServer configurations should be strictly adhered to. Using,
+   for example, `c.CylcUIServer.user-authorisation` rather than
+   `c.CylcUIServer.user-authorization` would result in cylc silently ignoring
+   this configuration.
+
+Methods of Assigning Permissions
+--------------------------------
+Assigning permissions can be done in two ways:
+
+ - using individual Cylc operations, e.g. ``play``, ``pause``
+ - using predefined access groups: ``READ``, ``CONTROL``, ``ALL``.
+
+Using both methods is supported, e.g ["READ", "stop", "pause"]
+
+
+Individual Operations
+^^^^^^^^^^^^^^^^^^^^^ 
+To assign users permissions, you can list the operations you wish to grant
+
+.. code-block:: python
+
+  c.CylcUIServer.user_authorization = {
+      "user1": ["read", "pause", "play"]
+      }
+
+Provided you have permission to give `user1` these permissions, as defined in the
+site configuration file, this will result in `user1` being able to see your
+workflows (from the `read` operation), and `pause` and `play` your workflows.
+
+
+Removing permissions can be achieved by prepending the operation with a ``!``.
+For example,
+
+.. code-block:: python
+
+   c.CylcUIServer.user_authorization = {
+       "group:groupA": ["read", "play", "stop"],
+       "user2": ["!stop"]
+   }
+
+Again, providing your site configuration permits you to grant this access,
+this configuration, for `user2` (who is a member of system group `groupA`) would
+result in them having "read" and "play" access. The the access to stop your workflows
+has been removed and this action will be forbidden.
+
+Access Groups
+^^^^^^^^^^^^^
+For convenience, cylc operations available in the ui have been bundled into
+access groups. These should be capitalized to distinguish from cylc operations.
+
+We currently support ``READ``, ``CONTROL`` and ``ALL`` and to remove permissions
+to operations in these groups, use ``!READ``, ``!CONTROL``, ``!ALL``.
+
+
+.. csv-table:: Access Group Mappings
+   :header: "Operation", "READ", "CONTROL", "ALL"
+
+   "Broadcast", , , "X"
+   "Ext-trigger",, "X", "X"
+   "Hold",, "X", "X"
+   "Kill",, "X", "X"
+   "Message",, "X", "X"
+   "Pause",, "X", "X"
+   "Ping","X", , "X"
+   "Play",, "X", "X"
+   "Poll",, "X", "X"
+   "Read","X", , "X"
+   "Release",, "X", "X"
+   "ReleaseHoldPoint",, "X", "X"
+   "Reload",, "X", "X"
+   "Remove",, "X", "X"
+   "Resume",, "X", "X"
+   "SetGraphWindowExtent",, "X", "X"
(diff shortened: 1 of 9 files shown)


src/user-guide/sharing-workflows/sharing-workflows.rst
```diff
@@ -0,0 +1,285 @@
...
+
+Interacting with Others Workflows
+---------------------------------
+If you have been granted access to another users workflows, you can view and
+interact with these workflows.
+Say, you, userA, wishes to interact with userB's workflows.
+You can do this by changing the URL `https://<hub>/user/userB`, using the hub
+of the userB. You should authenticate as yourself (userA) and, provided you
+have the correct permissions, you will see userB's workflows for interaction.
+
```
datamel, line 269: Not sure if screenshots would be appropriate here, so users know what to expect?

src/user-guide/sharing-workflows/sharing-workflows.rst
```diff
@@ -0,0 +1,285 @@
+.. _Authorization:
+
+Authorization
+=============
+
+Cylc supports multi user access. You can grant control of your
```
oliver-sanders, line 6: Would be good to clarify the layer of the system this comes from to set expectations (e.g. ATM the CLI doesn't support this):

```suggestion
The Cylc UI Server supports multi user access. You can grant control of your
```

src/user-guide/sharing-workflows/sharing-workflows.rst
```diff
@@ -0,0 +1,285 @@
...
+- ``username`` to indicate a specific user.
+
+All permissions are additive, if user appears elsewhere in configuration, for
+example as a member of a system group, the permission level is taken as the
+greatest possible.
+
+However, any negations are applied with priority.
+
+Note, default in site config does not contribute to permissions, this is used
+in the case where a user does not appear in the `c.CylcUIServer.user-authorization`.
```
oliver-sanders, line 37: Should this be an underscore rather than a hyphen?

src/user-guide/sharing-workflows/sharing-workflows.rst
```diff
@@ -0,0 +1,285 @@
...
+
+.. note::
+
+   Spellings of UIServer configurations should be strictly adhered to. Using,
+   for example, `c.CylcUIServer.user-authorisation` rather than
+   `c.CylcUIServer.user-authorization` would result in cylc silently ignoring
+   this configuration.
```
oliver-sanders, line 44: I think there is now some form of warning message for invalid configs in Jupyter Server. Will have to try it out to confirm.

src/user-guide/sharing-workflows/sharing-workflows.rst
```diff
@@ -0,0 +1,285 @@
...
+
+Individual Operations
+^^^^^^^^^^^^^^^^^^^^^ 
+To assign users permissions, you can list the operations you wish to grant
+
+.. code-block:: python
+
+  c.CylcUIServer.user_authorization = {
+      "user1": ["read", "pause", "play"]
+      }
```
oliver-sanders, line 64:
```suggestion
   c.CylcUIServer.user_authorization = {
       "user1": ["read", "pause", "play"]
   }
```

src/user-guide/sharing-workflows/sharing-workflows.rst
```diff
@@ -0,0 +1,285 @@
...
+
+Provided you have permission to give `user1` these permissions, as defined in the
+site configuration file, this will result in `user1` being able to see your
+workflows (from the `read` operation), and `pause` and `play` your workflows.
```
oliver-sanders, line 68: Single backquotes have meaning in RST, can use double backquotes for literals or regular quotes otherwise.

Will need to check use of single backquotes throughout this page.

src/user-guide/sharing-workflows/sharing-workflows.rst
```diff
@@ -0,0 +1,285 @@
...
+   "SetOutputs",, "X", "X"
+   "SetVerbosity",, "X", "X"
+   "Stop",, "X", "X"
+   "Trigger",, "X", "X"
+
+
+.. note::
+
+   ``READ`` access group (which exands to ``["read", "ping"]``) is a method for
+   granting non-interactive access.
```
oliver-sanders, line 124: ?

src/user-guide/sharing-workflows/sharing-workflows.rst
```diff
@@ -0,0 +1,285 @@
...
+An example user configuration:
+
+.. code-block:: python
+
+   c.CylcUIServer.user_authorization = {
+       "*": ["READ"],   
+       "group:groupA": ["CONTROL"],
+       "user1": ["read", "pause", "!play"],
+       "user2": ["!ALL"]
+       }
```
oliver-sanders, line 149:
```suggestion
   c.CylcUIServer.user_authorization = {
       "*": ["READ"],   
       "group:groupA": ["CONTROL"],
       "user1": ["read", "pause", "!play"],
       "user2": ["!ALL"]
   }
```

src/user-guide/sharing-workflows/sharing-workflows.rst
```diff
@@ -0,0 +1,285 @@
...
+  workflows, even if `user1` is a member of the system `groupA`. This is due to
+  negations taking precedence over additions.
+
+- ``"user2"`` is not permitted to view workflows, or perform any operations. 
+
+.. _site_configuration:
+
+Site Authorization Configuration
+--------------------------------
+`c.CylcUIServer.site-authorization`, which is loaded from
```
oliver-sanders, line 171:
```suggestion
``c.CylcUIServer.site-authorization``, which is loaded from
```

src/user-guide/sharing-workflows/sharing-workflows.rst
```diff
@@ -0,0 +1,285 @@
...
+- ``"*"``  represents any authenticated user. They have permission to view all
+  workflows and ping them on the gui.
+
+- ``"group:groupA"`` applies ``CONTROL`` permissions to any member of system
+  `groupA`.
+  Note that, since permissions are additive, these users will gain ``READ`` access
+  from the ``"*":["READ"]`` assignment.
+
+- ``"user1"`` will have permission to view workflows, ``pause`` but not ``play``
+  workflows, even if `user1` is a member of the system `groupA`. This is due to
```
oliver-sanders, line 162: "user1" is in double quotes in double backquotes on one line then in single backquotes on the next?

src/user-guide/sharing-workflows/sharing-workflows.rst
```diff
@@ -0,0 +1,285 @@
...
+  negations taking precedence over additions.
+
+- ``"user2"`` is not permitted to view workflows, or perform any operations. 
+
+.. _site_configuration:
+
+Site Authorization Configuration
```
oliver-sanders, line 169: We need to add a note somewhere to explain the limitations of this security model to make it clear that perceived weaknesses in the site-config are not shortcomings in Cylc.

Something along the lines of:

> The `site_authorization` configuration allows sites to configure sensible defaults and limits for the permissions users can delegate.
>
> Note, however, that as the UIS runs as the user they have full control over this process so, in theory, may bypass these restrictions in a variety of ways. As an extreme example a user can pass their credentials on to another person. Such behaviours cannot be controlled through technical measures so must be enforced by site policy.

src/user-guide/sharing-workflows/sharing-workflows.rst
```diff
@@ -0,0 +1,285 @@
...
+^^^^^^^^^^^^^^^^^^^^^^
+Changing authorization permissions in your `jupyter_config.py` will require the
+UI Server to be restarted before any changes are applied.
+
+Interacting with Others Workflows
+---------------------------------
+If you have been granted access to another users workflows, you can view and
+interact with these workflows.
+Say, you, userA, wishes to interact with userB's workflows.
+You can do this by changing the URL `https://<hub>/user/userB`, using the hub
```
oliver-sanders, line 266:
```suggestion
You can do this by changing the URL ``https://<hub>/user/userB``, using the hub
```

src/user-guide/sharing-workflows/sharing-workflows.rst
datamel: Ah yes, I remember you mentioning this. I couldn't spot any message but entirely possible I have missed them, unless it is a feature we have to turn on.

src/user-guide/sharing-workflows/sharing-workflows.rst
datamel: As in, with READ, you are providing view-only type access to other users. They will be unable to interact with workflows. 
I can remove this if you think it is obvious enough, through the naming of the access group.

src/user-guide/sharing-workflows/sharing-workflows.rst
datamel: I shall sort this out now.

src/user-guide/sharing-workflows/sharing-workflows.rst
oliver-sanders: It's the non-interactive access bit that's confusing me.

src/user-guide/sharing-workflows/sharing-workflows.rst
oliver-sanders: Here's my test:

`~/.cylc/hub/jupyter_config.py`:
```python
c.CylcUIServer.not_a_trait = False
```

```console
$ cylc gui
[W 2021-10-05 13:10:22.435 CylcUIServer] Config option `not_a_trait` not recognized by `CylcUIServer`.
...
```

datamel: Ah I have gone through the hub and couldn't spot a warning :) I shall change the docs.

datamel: Also, I wonder if there is a way of returning a warning if using `cylc hub`?

datamel:
Rebased. There is still an unresolved comment relating to this sentence...
` By the same token, however, because Cylc UI Servers (and Schedulers) run as the user, one user cannot compromise another user's account using Cylc.`
Any strong thoughts @hjoliver?
I have attempted to rewrite with little success, how about something like...
`By the same token, however, since these processes (Cylc UI Servers and Schedulers) run as the owner, granting a user access has no security implications for *their* account.
`

oliver-sanders (approved):

hjoliver:
Thanks @datamel I'll get this merged. Will reconsider the hard-to-word sentence as a follow-up.

PR #307, opened 2021-10-11
Testing

wxtim:
(It's still not passing, but the change to fix the other break is in metronnie 's PR https://github.com/cylc/cylc-doc/pull/304 - Somehow a `:ref:backCompat` (added in Ronnie's PR) has snuck into cylc-flow master...


MetRonnie:
@wxtim I would merge this but looks like it contains leakage from #305. If it's alright I'll push to master a hotfix commit node version -> 16 for test.yml

PR #330, opened 2021-11-23
Minor correction to DB diagram

MetRonnie:
Fix relationship between `task_pool` table and `task_prerequisites` table


src/ext/database_diagram.py
```diff
@@ -56,6 +56,8 @@ def get_relationships():
          CylcWorkflowDAO.TABLE_TASK_LATE_FLAGS],
         [CylcWorkflowDAO.TABLE_TASK_POOL, ONE_TO_MANY,
          CylcWorkflowDAO.TABLE_TASK_OUTPUTS],
+        [CylcWorkflowDAO.TABLE_TASK_POOL, ONE_TO_MANY,
+         CylcWorkflowDAO.TABLE_TASK_PREREQUISITES],
```
MetRonnie, line 60: Not sure if this is one-to-one or one-to-many. I went for the latter, as the same task (name, cycle point) can appear multiple times for different prerequisites

hjoliver (approved):

PR #342, opened 2021-11-24
8.0b3 installation changes

oliver-sanders:
* Update installation instructions.    
* Document Conda Forge package outputs.        
* Pin documented Conda environments to the Cylc "release" (i.e. minor        
  version).        
* Update test workflow to create the new environments.        
* Add environment (and with it a test) for mixed Conda-Forge / Anaconda        
  installations.     

Note: the conda tests are expected to fail at this stage, see the comment below and a link to a run where they pass.

PR #349, opened 2021-11-29
Tweak release messaging.

hjoliver:
I had a pending review of #342 that I forgot to finish off before it got trumped by other reviewers.

My main issue was:
> It's important to convey that although 8.0b3 is a still a "pre-release", it is now close enough to complete to be fully useable, with the disclaimer that it has not yet been heavily tested by users (although it has by us). Because, as recently reported on the forum, some sites are already facing new platforms without Python 2 available and our "not recommended for production use" messaging is putting them off Cylc 8.

I've attempted to strike a compromise that let's users decide if they want to take the risk, and that "Cylc 7 is stable and still available if needed".


hjoliver:
(Looks like the automatic replacement of `|release|` is not file-specific, so this shouldn't break that).

oliver-sanders (approved):

PR #357, opened 2021-12-02
deploy wf tweak

hjoliver:
The "create conda envs" action in the `deploy` workflow isn't working. 

It fails because of a path problem, and `continue-on-error` isn't working either (`unexpected value: 'true'`)

I couldn't figure this out quickly so I deleted the whole block, and used this branch to deploy the 8.0b3 docs.

Probably want to fix it rather than merge this.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -89,12 +89,6 @@ jobs:
           git clone --depth 1 --branch "${meto_rose}" https://github.com/metomi/rose ../rose
           pip install -e ../rose[all]
 
-      - name: create conda envs
-        uses: ./.github/actions/create-conda-envs
-        continue-on-error: ${{ github.event.inputs.skip_conda_environment_check }}
-        with:
-          python_version: '3.7'
-
       - name: checkout gh-pages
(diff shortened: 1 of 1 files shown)


hjoliver:
```
0s
Error: Can't find 'action.yml', 'action.yaml' or 'Dockerfile' under '/home/runner/work/cylc-doc/cylc-doc/.github/actions/create-conda-envs'.
  Did you forget to run actions/checkout before running your local action?
Error: .github/workflows/deploy.yml (Line: 94, Col: 28): Unexpected value 'true'
Error: The step failed and an error occurred when attempting to determine whether to continue on error.
Error: The template is not valid. .github/workflows/deploy.yml (Line: 94, Col: 28): Unexpected value 'true'
```

hjoliver:
The `undeploy` workflow is also broken.

MetRonnie:
>and `continue-on-error` isn't working either (`unexpected value: 'true'`)

This could be https://github.com/actions/runner/issues/1483

MetRonnie:
Closing as superseded by #358

PR #363, opened 2021-12-07
Tidy up file format documentation.

hjoliver:
Had a couple of new users confused about the file format. E.g., this was only documented in the tutorial:

>    We only use one set of square brackets at each level when writing nested sections on one line. But in the file, each nested sub-section gets a another set of square brackets as shown above.

So I added a new Reference section, and just referred out to that in the User Guide.
 
@wxtim may be some overlap with your tutorial rewrite (I just added a TODO comment to the tutorial docs here, since we haven't decided where that'll end up yet).

### DIFF
--- a/src/reference/config/file-format.rst
+++ b/src/reference/config/file-format.rst
@@ -6,6 +6,7 @@ The .cylc File Format
 Cylc global and workflow configuration files are written in a nested
 `INI`_-based format.
 
+.. _syntax:
 
 Syntax
 ------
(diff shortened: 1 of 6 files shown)


hjoliver:
One review will do.

wxtim (approved):

wxtim:
@hjoliver  - Approved for content. Feel free to merge after fixing build fail.

src/reference/config/file-format.rst
```diff
@@ -0,0 +1,179 @@
+.. _file-format:
+
+The .cylc File Format
+=====================
+
+Cylc global and workflow configuration files are written in a nested
+`INI`_-based format.
+
+
```
wxtim, line 9:
```suggestion
.. _syntax:

```

PR #366, opened 2021-12-10
Update notes about disallowed dir names in workflow name

MetRonnie:
Companion to https://github.com/cylc/cylc-flow/pull/4549

### DIFF
--- a/src/user-guide/installing-workflows.rst
+++ b/src/user-guide/installing-workflows.rst
@@ -67,8 +67,8 @@ addition of the option ``--directory=PATH/TO/SOURCE/DIRECTORY`` (alternatively,
 .. note::
 
    To avoid confusion, ``cylc install`` will fail if the workflow name
-   contains ``runN`` or ``run<number>``. These directory names are
-   reserved for the automatic numbering of run directories.
+   contains directory names that are reserved filenames, such as
+   ``log``, ``work``, ``runN``, ``run<number>``, ``_cylc-install`` etc.
 
 Configurable Source Directories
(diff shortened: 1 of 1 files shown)


src/user-guide/installing-workflows.rst
```diff
@@ -67,8 +67,8 @@ addition of the option ``--directory=PATH/TO/SOURCE/DIRECTORY`` (alternatively,
 .. note::
 
    To avoid confusion, ``cylc install`` will fail if the workflow name
-   contains ``runN`` or ``run<number>``. These directory names are
-   reserved for the automatic numbering of run directories.
+   contains directory names that are reserved filenames, such as
+   ``log``, ``work``, ``runN``, ``run<number>``, ``_cylc-install`` etc.
```
datamel, line 71: Wonder if we should put a full list of forbidden names rather than an etc.. Just for user clarity?

MetRonnie: Best I can do is
```rst
   .. autoattribute:: cylc.flow.workflow_files.WorkflowFiles.RESERVED_NAMES
```
But it isn't especially neat, nor does it include `run<number>`

image

datamel: Yes, I agree it isn't especially neat or elegant with the run number not included :(
Although personally think it best to be comprehensive in documentation, so if it were up to me I'd take the auto linked code. I'll leave it up to you @MetRonnie/second reviewer to decide.

MetRonnie: Done

datamel (approved):
Thanks @MetRonnie, checked and built.

PR #368, opened 2021-12-14
Create custom 404 page

MetRonnie:
Detects the version you tried to navigate to, offers a link to the docs homepage for that version. This is better than the current 404 page which offers no options.

To see a demo
- Go to https://metronnie.github.io/cylc-doc/8.0b3/html/7-to-8/index.html
- Try changing the version to 7.8.8 in the lower left sidebar

image

Note if you try entering an invalid version in the URL, it will only offer the link to the stable docs homepage

PR #373, opened 2022-01-04
Prevent 404.html from showing up in sidebar versions

MetRonnie:
Followup to #368

image

### DIFF
--- a/bin/version
+++ b/bin/version
@@ -34,7 +34,7 @@ DUMMY_FORMATS = [
 
 def get_versions(builddir):
     """Yield paths to each documented version."""
-    yield from pathlib.Path(builddir).glob('[0-9]*.*')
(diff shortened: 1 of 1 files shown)


bin/version
```diff
@@ -34,7 +34,7 @@ DUMMY_FORMATS = [
 
 def get_versions(builddir):
     """Yield paths to each documented version."""
-    yield from pathlib.Path(builddir).glob('[0-9]*.*')
+    yield from pathlib.Path(builddir).glob('[0-9]*.[0-9]*')
```
datamel, line 37: Haven't manually tested this. Will this yield the alphas/betas correctly?

MetRonnie: Seeing as we have the convention of always specifying the minor/2nd number, this should do it. Will check with Oliver

oliver-sanders: Think that's ok.

datamel (approved):

PR #376, opened 2022-01-10
Caution note for platform select by subshell added.

### LINKED ISSUE #374: explaining what happens if the [platform selection] command exits with a non-zero return code
At https://github.com/cylc/cylc-doc/pull/370/files#r772227988 @oliver-sanders  suggested:

> Adding a paragraph explaining what happens if the [platform selection] command exits with a non-zero return code.

wxtim:
Closes #374

### DIFF
--- a/src/user-guide/writing-workflows/runtime.rst
+++ b/src/user-guide/writing-workflows/runtime.rst
@@ -477,6 +477,12 @@ For example:
 Job hosts are always selected dynamically, for the chosen platform or
 platform group.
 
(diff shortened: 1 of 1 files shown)


oliver-sanders (approved):

oliver-sanders:
I think if you have `submit-retried` configured the subshell eval will be repeated (until you run out of retries).

If so should add that so it doesn't sound like this logic bypasses the retry logic.

PR #377, opened 2022-01-14
universal id

oliver-sanders:


wxtim (approved):
- [x] Read
- [x] Built
- [x] Double checked `cylc-graph-reversible.svg`.
- [x] Grepped for old style IDs.

PR #379, opened 2022-01-19
documentation for cylc config --platform-*

wxtim:
Sibling of https://github.com/cylc/cylc-flow/pull/4576

### DIFF
--- a/src/7-to-8/cheat-sheet.rst
+++ b/src/7-to-8/cheat-sheet.rst
@@ -177,22 +177,6 @@ View the parsed workflow configuration:
 
          cylc config <name/path>
 
-
-List configured platforms and platform groups
----------------------------------------------
-
-.. seealso::
-
-   :ref:`ListingAvailablePlatforms`
-
-
-``cylc config --platform-names``
-   Print a list of configured platforms and platform groups.
-
-``cylc config --platform-meta``
-   Print out the full configuration of ``[platfroms]`` and
-   ``[platform groups]``
-
 Opening User Interfaces
 -----------------------
 
--- a/src/7-to-8/cheat-sheet.rst
+++ b/src/7-to-8/cheat-sheet.rst
@@ -177,6 +177,22 @@ View the parsed workflow configuration:
 
          cylc config <name/path>
 
+
+List configured platforms and platform groups
+---------------------------------------------
+
+.. seealso::
+
+   :ref:`ListingAvailablePlatforms`
+
+
+``cylc config --platform-names``
+   Print a list of configured platforms and platform groups.
+
+``cylc config --platform-meta``
+   Print out the full configuration of ``[platfroms]`` and
+   ``[platform groups]``
+
 Opening User Interfaces
 -----------------------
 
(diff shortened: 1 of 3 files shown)


src/7-to-8/cheat-sheet.rst
```diff
@@ -177,6 +177,22 @@ View the parsed workflow configuration:
...
+   :ref:`ListingAvailablePlatforms`
+
+
+``cylc config --platform-names``
+   Print a list of configured platforms and platform groups.
+
+``cylc config --platform-meta``
+   Print out the full configuration of ``[platfroms]`` and
+   ``[platform groups]``
+
```
oliver-sanders, line 195: This page is for Cylc 7 -> 8 CLI migration.

wxtim: I thought that wanting to know what platforms were available is exactly the sort of thing someone new to Cylc 8 from Cylc 7 might want a  quick reference to. But happy to remove if this is just for things which have a clearer mapping.

oliver-sanders: The purpose of the migration guide is to outline the changes people must make to transition to Cylc 8. We need to keep it as minimal as possible and "signpost" into the main documentation. If we listed every helpful command this would end up pages long.

wxtim, line 195:
```suggestion
```

oliver-sanders (approved):

src/reference/config/writing-platform-configs.rst
```diff
@@ -13,16 +13,21 @@ Writing Platform Configurations
...
+To see a list of available platforms::
+
+   cylc config --platform-names
 
-To see available platforms::
+To see the full configuration of available platforms::
 
-   cylc config -i [platforms]
-   cylc config -i [platform groups]
+   cylc config --platform-meta
```
hjoliver, line 30: Note https://github.com/cylc/cylc-flow/pull/4576#discussion_r791218359

(Change to `--platforms`?).

And it wouldn't hurt to mention that this option is just shorthand for `cylc config -i 'platforms' -i 'platform groups'`.

hjoliver (changes requested):

src/reference/config/writing-platform-configs.rst
wxtim:
```suggestion
   cylc config --platform-meta
   
This is equivalent to `cylc config -i 'platforms' -i 'platform groups'`
```

wxtim: @hjoliver  I have added a comment about platform meta being the same as `cylc config -i 'platforms' -i 'platform groups'`. I deliberately avoided using `--platforms` because I wasn't sure which of the above it should map to.

hjoliver: (My point was, in the cylc-flow PR, `--platform-meta` is an incorrect name because it prints the full platforms config, not just the metadata).

wxtim: Ok, Reasonable

wxtim:
@MetRonnie , I think that I've responded to Hilary's comments and you could merge if you agree that I have.

oliver-sanders (approved):

hjoliver (approved):

PR #381, opened 2022-01-20
make: Add sphinx-autobuild support

oliver-sanders:
Makes working with Sphinx much, much easier.

```console
$ make watch
```

Once you do this the docs will rebuild (incrementally where possible) whenever. Your browser will automatically refresh when this happens 🎉 😁 🥳 🕺!

There is also a `make watch-cylc` target which listens for changes in the `cylc-flow`, `cylc-uiserver` & `cylc-rose` repos. Note this performs a non-incremental build which is substantially slower.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,11 +22,33 @@ clean:
 	rm -rf src/plugins/install/built-in
 	rm -rf src/user-guide/task-implementation/job-runner-handlers
 
+watch: clean
+	# rebuild incrementally whenever documentation files are changed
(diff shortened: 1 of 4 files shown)


oliver-sanders:
(haven't tested it on NFS, sadly it might not work there)

MetRonnie (approved):
Tested on NFS

oliver-sanders:
Great! Might be worth investigating what library they are using to detect filesystem events at some point (NFS is tricky for this stuff). If it's efficient it could provide us with a more efficient, incremental `cylc scan` which would be great.

hjoliver (approved):
Nice.

PR #392, opened 2022-02-07
Update script item style reccommendations

### LINKED ISSUE #149: Update script item indentation advice.
We originally recommended indenting the internal lines of multi-line script strings to the left margin, because they are verbatim bash job script inclusions and not part of the surrounding config file syntax:
```
[runtime]
   [[cheese]]
       script = """
echo HELLO
echo GOODBYE
"""
```

However, automatic dedent processing now means it is safe to do this  (even for heredocs - athough you have to be careful to get the indentation exactly right for the heredoc end marker):
```
[runtime]
   [[cheese]]
       script = """
           echo HELLO
           echo GOODBYE
       """
```
... which is generally considered to look better (even though it is technically wrong :stuck_out_tongue_winking_eye: )

Given that, we should:
- [ ] update indentation advice in the Suite Design Guide
- [ ] explain why it is OK to indent like this - for "the thinking user" :brain:

MetRonnie:
Closes #149

- [x] update indentation advice in the Suite Design Guide
- [x] explain why it is OK to indent like this - for "the thinking user" 🧠

### DIFF
--- a/src/workflow-design-guide/style-guide.rst
+++ b/src/workflow-design-guide/style-guide.rst
@@ -92,45 +92,49 @@ to, and trailing comments should be preceded by two spaces, as shown above.
 Script String Lines
 ^^^^^^^^^^^^^^^^^^^
 
-Script strings are written verbatim to task job scripts so they should really
-be indented from the left margin:
+Script strings are written verbatim to task job scripts.
+
+.. code-block:: cylc
+
+   [runtime]
+       [[foo]]
+           script = echo "Hello, Mr. Thompson"
+
+If using a triple-quoted string, any common leading whitespace is trimmed
+using the logic of
+`Python's textwrap.dedent <https://docs.python.org/2/library/textwrap.html#textwrap.dedent>`_.
+As such, it is recommended to indent like any other triple-quoted string
+setting in Cylc:
 
 .. code-block:: cylc
 
    [runtime]
        [[foo]]
            # Recommended.
-           post-script = """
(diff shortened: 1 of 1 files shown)


src/workflow-design-guide/style-guide.rst
```diff
@@ -92,45 +92,49 @@ to, and trailing comments should be preceded by two spaces, as shown above.
...
+
+.. code-block:: cylc
+
+   [runtime]
+       [[foo]]
+           script = echo "Hello, Mr. Thompson"
+
+If using a triple-quoted string, any common leading whitespace is trimmed
+using the logic of
+`Python's textwrap.dedent <https://docs.python.org/2/library/textwrap.html#textwrap.dedent>`_.
```
oliver-sanders, line 105: We have hooked up the Python docs via intersphinx so you can use something like:

```rst
:py:func:`textwrap.dedent`
```

To reference `dedent` in the Python 3 docs.

src/workflow-design-guide/style-guide.rst
```diff
@@ -143,8 +147,7 @@ have many levels of indentations.
                _EOF_
```
oliver-sanders, line 147: Is this EOF aligned with the `cat`?

oliver-sanders (commented):
Looks good.

src/workflow-design-guide/style-guide.rst
MetRonnie: yes

oliver-sanders (approved):

PR #397, opened 2022-02-09
Spellcheck

wxtim:
Add sphinxcontrib-spellcheck.
The problem is that despite configuring it to use `lang='en_GB'` this doesn't seem to be all that reliable and tends to object to non US English.
Having run the spell check recently, however, I think it's worth putting this up, despite its weaknesses.

Further work - we should decide which of the two valid spelling of "parameteri[sz]e" we want to use. Both are legal in UK English, although "z" is a bit archaic.


src/workflow-design-guide/portable-workflows.rst
```diff
@@ -263,7 +263,7 @@ workflow config file to automatically set site-specific workflow inputs and ther
 avoid the need to make manual changes every time you check out and run a new
 version. The site switch itself has to be set of course, but there may be other
 settings too such as model parameters for a standard local test domain. Just
-put these settings in ``opt/rose-suite-niwa.conf`` (for site "niwa")
+put these settings in ``opt/rose-suite-niwa.conf`` (for site "NIWA")
```
oliver-sanders, line 266:
```suggestion
put these settings in ``opt/rose-suite-niwa.conf`` (for site ``niwa``)
```

src/spelling_wordlist.txt
```diff
@@ -0,0 +1,249 @@
+GScan
+acknowledgement
+aldergrove
+analyse
+aren
```
MetRonnie, line 5: Does it not handle contractions with apostrophes? e.g. `aren't`

src/spelling_wordlist.txt
```diff
@@ -0,0 +1,249 @@
...
+arg
+args
+auth
+baz
+behaviour
```
MetRonnie, line 10: If `en_GB` works, should be able to take out these

src/spelling_wordlist.txt
```diff
@@ -0,0 +1,249 @@
...
+boolean
```
MetRonnie, line 11: Strange that this isn't recognised. Is it because it's technically a proper noun so ought to be capitalised?

wxtim: OED says capitalized...

src/spelling_wordlist.txt
wxtim: No, annoyingly.

wxtim:
@oliver-sanders  @MetRonnie  
It looks like Pyenchant is a dangerously under-supported library (https://github.com/pyenchant/pyenchant/issues/93) - and getting it to pass is becoming tricky. I can just pull the words it doesn't like off GHA and fix it manually, but I'm not sure I like the level of faffing.

Happy to
1) Close
2) Just accept that we might have a lot of exceptions in the custom dict and use it to spot future spelling fialures.

MetRonnie:
Hmm, probably not worth the effort then

wxtim:
> Hmm, probably not worth the effort then

All option 2 requires is piping the output of the `make spelling` to a well constructed awk command and then quickly checking the output of the awk command for obviously wrong stuff.

We then have a working setup which will object to new typos.

wxtim:
Not worth it for reasons described above.

PR #403, opened 2022-02-11
Update cylc clean docs

MetRonnie:
Accompanies #4675


src/user-guide/removing-workflows.rst
```diff
@@ -110,12 +104,13 @@ A note on symlink directories
...
-manager), only the symlinks themselves will be deleted, not the actual targets.
-In contrast, ``cylc clean`` follows the symlinks and deletes the targets.
+If you manually delete a run directory (e.g., using the ``rm`` command or the
+file manager), only the symlinks themselves will be deleted, not the actual
+targets. In contrast, ``cylc clean`` follows the symlinks and deletes the
+targets.
 
-- It does this for the symlinks that can be set in
-  :cylc:conf:`global.cylc[install][symlink dirs]` only, not any custom
+- It does this only for the symlinks that can be set in
```
datamel, line 112: Take or leave, I found myself re-reading this sentence, although perhaps personal preference.
```suggestion
- It only does this for the symlinks that are set in
```

datamel (approved):
I have built the docs and read the changes. Would it be worth mentioning use case for `--yes`?

MetRonnie:
> I have built the docs and read the changes. Would it be worth mentioning use case for `--yes`?

I thought about it, but I don't think we need to advertise it any more than `--help` offers (and anyone running `cylc clean` in a non-interactive session will get an error message saying use `--yes` anyway)

MetRonnie:
Test failure is due to setup.py falling foul of new flake8-simpilify rule, not worth fixing, will open a PR to convert setup.py to setup.cfg anyway

oliver-sanders (approved):

PR #405, opened 2022-02-14
Convert to setup.cfg / PEP 517

MetRonnie:


pyproject.toml
```diff
@@ -0,0 +1,7 @@
+[build-system]
+# NOTE: `pip install build` to build with `python -m build`
+requires = [
+  "setuptools >= 40.9.0",
+  "wheel"
+]
+build-backend = "setuptools.build_meta"
```
MetRonnie, line 7: Actually, because we're not publishing this as a package, do we even need `pyproject.toml`?

oliver-sanders: No.

oliver-sanders (approved):

pyproject.toml
MetRonnie: Deleted it and force-pushed. (Had a go building the project beforehand and it built successfully, to lend credibility to the setup.cfg implementation)

MetRonnie: Further update: looks like we don't need any _content_ in `pyproject.toml` if we're not publishing the package, so force pushed empty `pyproject.toml`

oliver-sanders (approved):

PR #407, opened 2022-02-17
Improve back-compat info.

### LINKED ISSUE #389: 728: include/exclude at start up
The Cylc 7 `include at startup` and `exclude at startup` configurations have been removed with no failsafe deprecation logic.

We should add an entry to the major changes section of the migration guide for them.

Will need to read up about the old configs and try them out as I think they were a bit strange from memory. Should try to offer contemporary solutions for implementing them.

hjoliver:
Close #389

Document obsoleted `include/exclude at start-up` workflow config.

Also fixes a few typos and adds a 7-8 migration warning to check for use of the CLI in task scripting.

### DIFF
--- a/src/7-to-8/major-changes/excluding-tasks.rst
+++ b/src/7-to-8/major-changes/excluding-tasks.rst
@@ -3,6 +3,18 @@
 Excluding Tasks at Start-up is Not Supported
 ============================================
 
+.. admonition:: Does This Change Affect Me?
+   :class: tip
+
+   This will affect you if your workflows use the following configurations:
+
+   * ``[scheduling][special tasks]include at start-up``
+   * ``[scheduling][special tasks]exclude at start-up``
+
+
+Overview
+--------
+
 The Cylc 7 sheduler allowed you to exclude tasks from the scheduler at start-up: 
 
 .. code-block:: cylc
--- a/src/7-to-8/major-changes/excluding-tasks.rst
+++ b/src/7-to-8/major-changes/excluding-tasks.rst
@@ -1,0 +1,31 @@
+.. _MajorChangesExcludingTasksAtStartup:
+
+Excluding Tasks at Start-up is Not Supported
+============================================
+
+The Cylc 7 sheduler allowed you to exclude tasks from the scheduler at start-up: 
+
+.. code-block:: cylc
+
+   # Cylc 7 only 
+   [scheduling]
+      [[special tasks]]
+           include at start-up = foo, bar, baz  # Cylc 8 ERROR!
+           exclude at start-up = bar  # Cylc 8 ERROR!
+
+The first config item above excludes all task names not in the include-list;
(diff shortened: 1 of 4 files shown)


hjoliver:
**NOTE** as per offline discussion, we actually could quite easily support this old feature as deprecated.  Suddenly obsoletion  seems somewhat at odds with our strategy generally, even if "exclude at start-up" was not much used.

If we change tack on this I can put up a cylc-flow PR and perform surgery on this one...

src/7-to-8/major-changes/excluding-tasks.rst
```diff
@@ -0,0 +1,31 @@
...
+
+The Cylc 7 scheduler started up with an instance of every task in its "task
+pool", and the workflow evolved by each task spawning its own next-cycle
+instance at the right time. So, if you excluded a task a start-up it would not
+run in the workflow at all unless manually inserted later at runtime.
+
+The Cylc 8 scheduler starts up with only the initial tasks in the graph and the 
+workflow evolves by spawning new tasks on demand as dictated by the graph.
+Consequently excluding a task at start up as described above would have no
+effect at all on most tasks.
```
hjoliver, line 27: Note ("as described above") - I've carefully avoided saying that it could not be done in Cylc 8.

src/7-to-8/major-changes/excluding-tasks.rst
```diff
@@ -0,0 +1,31 @@
+.. _MajorChangesExcludingTasksAtStartup:
+
+Excluding Tasks at Start-up is Not Supported
+============================================
+
```
oliver-sanders, line 5: Should add a "does this change affect me" box.

src/7-to-8/major-changes/excluding-tasks.rst
```diff
@@ -0,0 +1,31 @@
...
+run in the workflow at all unless manually inserted later at runtime.
+
+The Cylc 8 scheduler starts up with only the initial tasks in the graph and the 
+workflow evolves by spawning new tasks on demand as dictated by the graph.
+Consequently excluding a task at start up as described above would have no
+effect at all on most tasks.
+
+This feature also predated the current Cylc dependency graph configuration. To
+exclude tasks now without entirely removing them from the workflow definition,
+just comment them out of the graph.
```
oliver-sanders, line 31: Could do with an example of how to include / exclude tasks at startup without this syntax i.e. `R1` and `<rec> ! R1`?

hjoliver: Hmm, that's not how it works though. (I'll comment offline).

oliver-sanders: Fine, probably hard to come up with an example for this.

oliver-sanders (approved):

wxtim (approved):

PR #409, opened 2022-02-21
Revert markupsafe thingies

MetRonnie:
No longer needed as we have pinned markupsafe to a max version in 8.0rc2.dev

PR #411, opened 2022-02-21
setup: markupsafe

oliver-sanders:
Having real fun deploying this one...

Temp patch, strip later.


MetRonnie (approved):

MetRonnie:
Revert added in #409

PR #425, opened 2022-03-14
Fix nightly

MetRonnie:
Fix nightly builds and bring back `nightly.CCYY-MM-DD` version name in sidebar because it makes it more obvious you are looking at the nightly docs and we can more easily notice when the build is failing as the date becomes out-of-date.

- [x] Relies on releasing cylc-sphinx-extensions > 1.3.4

### DIFF
--- a/src/reference/environments/envs/cylc-all.yml
+++ b/src/reference/environments/envs/cylc-all.yml
@@ -1,8 +1,8 @@
-name: cylc-|release|
+name: cylc-|version|
 channels:
   - conda-forge
 dependencies:
-  - cylc-flow =|release|
+  - cylc-flow =|version|
   - cylc-uiserver
   - cylc-rose
   - metomi-rose
--- a/.github/actions/create-conda-envs/action.yml
+++ b/.github/actions/create-conda-envs/action.yml
@@ -39,7 +39,7 @@ runs:
           # TODO: add explicit python dep
           sed \
             -e 's/python =.*/python =${{ inputs.python_version }}/' \
-            -e "s/|release|/${CYLC_RELEASE}/" \
+            -e "s/|version|/${CYLC_RELEASE}/" \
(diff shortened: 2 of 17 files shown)


src/reference/environments/envs/cylc-all.yml
```diff
@@ -1,8 +1,8 @@
-name: cylc-|release|
+name: cylc-|version|
 channels:
   - conda-forge
 dependencies:
-  - cylc-flow =|release|
+  - cylc-flow =|version|
```
oliver-sanders, line 5: Not sure why you've changes these, they are example recipes so `cylc-flow` should be pinned to the minor version, i.e. the `|release|` not the `|version|`.

MetRonnie: I was trying to make these consistent with what is noted here: https://github.com/cylc/cylc-doc/blob/34f8c5539652b8a9da786fa4febb5d5a7e58189f/src/conf.py#L93-L96

oliver-sanders: That's because Sphinx defines release/version the other way around to normal hence this line:

> # NOTE: Sphinx considers version/release the other way around to us

MetRonnie: But the point is `version` should have been what we were using for pinning, not `release`

MetRonnie:
Note: at least, the fix to the deploy workflow is needed for 8.0rc2

src/reference/environments/envs/cylc-all.yml
oliver-sanders: Ok, tested with 8.0.1, it came out in the environment files as 8.0 so good. I find "version" meaning the short version confusing, but can deal with it.

oliver-sanders (approved):

PR #427, opened 2022-03-16
Redo the User Guide Reflow section.

hjoliver:
To be merged AFTER https://github.com/cylc/cylc-flow/pull/4738 and https://github.com/cylc/cylc-flow/pull/4739

I spent a fair amount of time trying to edit this down to something simple and clear ~~but recent history shows it is difficult to describe this stuff in universally agreed terms, so I won't bother adding extra cross-referencing etc. just yet.~~ [it seems we have largely agreed].

Note this re-does the original Reflow section, which is now (well, about to be) out of date. In addition to this we may need, earlier in the guide, something on plain old triggering that mainly describes the default case, but refers forward to this Multiple Flows section for the full story.


hjoliver:
(Last update re-orders the somewhat randomly order of sub-sections in UG "Running Workflows")

src/glossary.rst
```diff
@@ -43,26 +43,36 @@ Glossary
...
       .. seealso::
 
          * :ref:`Cylc User Guide <TaskRetries>`
 
 
+   submit number
+   task submit number
+      Every time a task re-runs, whether by automatic :term:`retry` or manual
+      triggering, its submit number increments. Submit number appears in the
+      job log path so that log files don't get overwritten.
```
dpmatthews, line 65: Suggest adding:
It is passed to the job environment as ``$CYLC_TASK_SUBMIT_NUMBER``.

dpmatthews, line 65: Suggest adding:
It is passed to the job environment as ``$CYLC_TASK_SUBMIT_NUMBER``.

src/user-guide/running-workflows/dynamic-behaviour.rst
```diff
@@ -3,19 +3,18 @@
...
+The ``cylc broadcast`` command overrides task :cylc:conf:`[runtime]`
+settings in a running scheduler. You can think of it as communicating
+changed settings (including information via environment variables) to selected
+upcoming tasks via the scheduler.
+
+Broadcasts can target all tasks, groups of tasks, or specific tasks, at all
+cycle points or at specific cycle points. See ``cylc broadcast --help`` for
+detailed information.
+
+Broadcast settings targeting a specific task ID or cycle point expire as the
```
dpmatthews, line 15: The task ID is irrelevant here isn't it (it may be a specific ID at all cycle points)?

dpmatthews, line 15: The task ID is irrelevant here isn't it (it may be a specific ID at all cycle points)?

src/user-guide/running-workflows/reflow.rst
```diff
@@ -1,86 +1,227 @@
...
-The new flow will regenerate and republish cycle 5 products before naturally
-coming to a halt, because the triggered tasks do not lead on to the next cycle
-and the rest of the graph.
+By default, manual triggering (with ``cylc trigger`` or the UI) starts a new
+:term:`front of activity<flow front>` in current flows.
+But it can also start new flows and trigger flow-independent single tasks.
 
-Meanwhile, the original flow will carry on unaffected, from cycle 8.
+In the diagrams below, the grey nodes represent tasks in flow ``1``, and
+the blue ones stem from a single manual triggering event.
```
dpmatthews, line 56: The blue ones can also represent tasks in flow 1.
How about:
The grey nodes represent tasks which run as a result of the original flow.
The blue nodes represent tasks which run as a result of the trigger.

dpmatthews, line 56: The blue ones can also represent tasks in flow 1.
How about:
The grey nodes represent tasks which run as a result of the original flow.
The blue nodes represent tasks which run as a result of the trigger.

src/user-guide/running-workflows/reflow.rst
```diff
@@ -1,86 +1,227 @@
...
+   active flow numbers. Subsequently, each of those flows will consider this
+   task to have run already.
+
+   **Ahead of active flows** this starts a new front of activity for the active
+   flows, which by default can continue on without waiting for catch up.
+
+   .. image:: ../../img/same-flow-n.png
+
+   With ``--wait``, the result is the same except that any action downstream of 
+   the triggered task is delayed until the first flow catches it.
```
dpmatthews, line 71: I think this needs a separate diagram - only the one task should be shown in blue?
I'm also not sure a single diagram really illustrates what's going on.
Could we use animates gif's?
Is there any way we could generate this diagrams?
Follow-up issue?

dpmatthews, line 71: I think this needs a separate diagram - only the one task should be shown in blue?
I'm also not sure a single diagram really illustrates what's going on.
Could we use animates gif's?
Is there any way we could generate this diagrams?
Follow-up issue?

src/user-guide/running-workflows/reflow.rst
hjoliver: That's what I meant (I didn't say the blue ones can't also be in flow 1) .. however, your suggestion is probably clearer.

src/user-guide/running-workflows/reflow.rst
```diff
@@ -1,86 +1,227 @@
...
+   .. image:: ../../img/no-flow-n.png
+
+Special Case: Triggering ``n=0`` Tasks
+   Tasks in the ``n=0`` window are :term:`active`, :term:`active-waiting`, or
+   :term:`incomplete`. Their flow membership is already determined - that of
+   the parent tasks that spawned them.
+
+   - Triggering an :term:`active task` has no effect (it is already triggered).
+   - Triggering an :term:`active-waiting task` runs it immediately in the same flow.
+   - Triggering an :term:`incomplete task` re-runs it immediately in the same flow.
```
dpmatthews, line 112: I understand from @oliver-sanders this isn't quite right.
Triggering an active-waiting task that's waiting on an internal queue runs it immediately.
In all other cases the trigger queues the task so it may require a second trigger to run immediately.

dpmatthews, line 112: I understand from @oliver-sanders this isn't quite right.
Triggering an active-waiting task that's waiting on an internal queue runs it immediately.
In all other cases the trigger queues the task so it may require a second trigger to run immediately.

src/user-guide/running-workflows/reflow.rst
```diff
@@ -1,86 +1,227 @@
...
+   You may need to manually stop the new flow once its job is done, to avoid
+   re-running more than you want to, if the new flow leads into the main
+   trunk of the graph.
+
+Rewinding a Workflow
+   To rewind the workflow to an earlier point, perhaps to regenerate data and/or 
+   allow the workflow to evolve a new path into the future, trigger a new
+   flow at the right place and then stop the original flow. (Alternatively,
+   stop the scheduler, install a new instance of the workflow, and play it
+   from the desired place in the graph).
```
dpmatthews, line 182: Installing a new instance isn't the same because you lose the contents of the `share` directory.
I'd suggest removing the bit  in brackets.

dpmatthews, line 182: Installing a new instance isn't the same because you lose the contents of the `share` directory.
I'd suggest removing the bit  in brackets.

src/user-guide/running-workflows/reflow.rst
```diff
@@ -1,86 +1,227 @@
...
+   You can trigger individual tasks as many times as you like with
+   ``--flow=none``, without affecting the workflow. The task :term:`submit
+   number` will increment each time.
+
+Processing Flow-Specific Data
+   Flow numbers are passed to job environments, so it is possible to have
+   different flows process different datasets through the same graph. However
+   **we do not recommend this**. That's what cycling is for; and besides, every
+   task would have to be capable of processing multiple datasets at once in
+   case of :ref:`flow-merging`.
```
dpmatthews, line 194: Why describe this if it's not a valid use case for this functionality?

dpmatthews, line 194: Why describe this if it's not a valid use case for this functionality?

hjoliver: It is a valid use-case, which I'm sure will occur to some users. It's just that you need to be very aware of the implications of flow merging.  So I think we should mention it, but warn about potential problems. I'll change the wording though.

src/user-guide/running-workflows/reflow.rst
hjoliver: Oops, good spot.  Yes, every task goes through the queue first.

src/user-guide/running-workflows/dynamic-behaviour.rst
hjoliver: No, task ID includes cycle point.  So you can broadcast to a specific task instance.

src/user-guide/running-workflows/reflow.rst
hjoliver: The diagram for this case was already in the repo, I just forgot to include it in the document. Fixed.

Follow-up issue for improvements: #457

hjoliver:
@dpmatthews - for some reason the GH UI has duplicated all of your comments, so I resolved one of each.

hjoliver:
Comments address; rebased and deconflicted.

dwsutherland (approved):
Finally got docs building in my system again.

Checked changes against browser results.. Looks good to me 👍

src/user-guide/running-workflows/dynamic-behaviour.rst
dpmatthews: Ah, OK. Got confused because you refer to "specific tasks" in the paragraph above.
Still now sure we really need to mention task IDs here though.
If we just say "settings targeting a specific cycle point" that would include task IDs wouldn't it?

hjoliver: Yes, fair enough. Especially as the broadcast CLI doesn't take task IDs as such (it takes cycle points and/or names, which can single out specific task IDs). I'll tweak the wording...

hjoliver: Done.

dpmatthews (approved):
Looks good

PR #410, opened 2022-02-21
deploy: fix workflow

oliver-sanders:

PR #434, opened 2022-03-22
actions: fix create conda envs

oliver-sanders:
Action will still fail, but now for a valid reason, 8.0rc2 isn't out yet.

This will block the cylc-doc deployment, however, there is a flag in the action to ignore this step so this isn't a true blocker to 8.0rc2 cylc-doc release.

(we can still run the workflow against 8.0rc1 etc by manually triggering it)

### DIFF
--- a/.github/actions/create-conda-envs/action.yml
+++ b/.github/actions/create-conda-envs/action.yml
@@ -17,7 +17,7 @@ runs:
         # get cylc version...
         if [[ -z '${{ inputs.cylc_flow_version }}' ]]; then
           # ... from cylc library
-          CYLC_RELEASE="$(python ./docs/src/lib/cylc_release.py)"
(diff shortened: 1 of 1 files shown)


.github/actions/create-conda-envs/action.yml
```diff
@@ -17,7 +17,7 @@ runs:
         # get cylc version...
         if [[ -z '${{ inputs.cylc_flow_version }}' ]]; then
           # ... from cylc library
-          CYLC_RELEASE="$(python ./docs/src/lib/cylc_release.py)"
+          CYLC_RELEASE="$(python ./src/lib/cylc_release.py)"
```
MetRonnie, line 20: This is a revert of #410

I think the proper fix will be to add an input to this action called something like `working_directory` that gives the path to `src`, and use that in this line

MetRonnie:
Will open an alternative PR

PR #459, opened 2022-04-19
Update docs for `--fcp`/`--stopcp`

MetRonnie:
Associated with https://github.com/cylc/cylc-flow/pull/4827

These docs were actually incorrect for `--final-cycle-point`, they were only correct for `--stop-cycle-point`. With this change, and after https://github.com/cylc/cylc-flow/pull/4827, they will be correct for both.

**Requirements check-list**


hjoliver (approved):

hjoliver:
The associated cylc-flow PR has been merged.

PR #460, opened 2022-04-21
Update docs re: validation

MetRonnie:
Accompanies cylc/cylc-flow#4829

**Requirements check-list**

### DIFF
--- a/src/tutorial/runtime/introduction.rst
+++ b/src/tutorial/runtime/introduction.rst
@@ -306,8 +306,7 @@ Validating A Workflow
 
 .. code-block:: console
 
-   $ cd ~/cylc-src/my_workflow
-   $ cylc validate .
+   $ cylc validate ~/cylc-src/my_workflow
 
 
 Installing A Workflow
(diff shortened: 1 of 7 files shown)


src/tutorial/runtime/introduction.rst
```diff
@@ -385,11 +384,11 @@ Numbered run directories
 
 .. code-block:: console
 
-   $ cylc validate my_workflow
+   $ cylc play my_workflow
```
oliver-sanders, line 387: Fine but what's the reason?

src/user-guide/writing-workflows/scheduling.rst
```diff
@@ -1790,8 +1790,6 @@ finish trigger, because the underlying member outputs must already be optional.
 Graph Branching
 ---------------
 
-.. versionadded:: 8.0.0
-
```
oliver-sanders, line 1794: Might be worth keeping this?

oliver-sanders (approved):

src/tutorial/runtime/introduction.rst
MetRonnie: Earlier on L309 we say run `cylc validate ~/cylc-src/my_workflow`, so I wanted to avoid any implication that `cylc validate my_workflow` is short for that

src/user-guide/writing-workflows/scheduling.rst
MetRonnie: But branching was still possible in Cylc 7? There is a note about the difference between 7 and 8 just below

oliver-sanders: Fair enough.

PR #464, opened 2022-05-04
Remote owner

oliver-sanders:
Added a page to the migration guide to cover the retirement of `[runtime][remote]owner` and the `--owner` CLI option.

Looks like we haven't documented the SSH approach to configuring remote usernames so chucked that in here too.

I'm not too sure about how reliably rsync will pick up on remote usernames. I presume this relies on SSH being configured as the transport layer as I wouldn't expect the RSH protocol to load the SSH config file? I've added a best-guess note advising RSync to be configured to use the SSH protocol when remote usernames are configured in this way. @dpmatthews any insights?

**Requirements check-list**


src/7-to-8/cheat-sheet.rst
```diff
@@ -137,14 +137,10 @@ Delete the workflow :term:`run directory` (leave source files untouched):
      - ::
 
          rose suite-clean <name>
-     - Recommended::
+     - ::
 
          cylc clean <name>
```
oliver-sanders, line 142: unrelated: `cylc clean` now works remotely so we can remove this

src/7-to-8/summary.rst
```diff
@@ -1,4 +1,4 @@
-.. _overview:
```
oliver-sanders, line 1: unrelated: vague label which doesn't seem to be used anywhere at present, this will most likely conflict (e.g. intersphinx) so swapping it out for something more specific.

src/7-to-8/major-changes/remote-owner.rst
```diff
@@ -0,0 +1,46 @@
...
+Cylc.
+
+.. warning::
+
+   GNU/BSD RSync is usually configured to use SSH for its transport layer,
+   however, other options, namely RSH may also be available.
+
+   If using the SSH config to set your remote username ensure RSync is configured
+   to use the SSH transport layer, or that the chosen transport layer is itself
+   appropriately configured.
```
dpmatthews, line 42: I'm not convinced we need this warning. rsync has used ssh by default for many years. Cylc assumes the use of ssh for contacting remote hosts so it's reasonable to assume that rsync will also be using ssh.

src/7-to-8/major-changes/remote-owner.rst
```diff
@@ -0,0 +1,46 @@
+Remote Usernames
+================
+
+.. admonition:: does this change affect me?
+   :class: tip
+
+   * If set task's ``[runtime][remote]owner`` configuration.
```
dpmatthews, line 7: Missing "you"?

src/7-to-8/major-changes/remote-owner.rst
```diff
@@ -0,0 +1,46 @@
...
+
+.. warning::
+
+   GNU/BSD RSync is usually configured to use SSH for its transport layer,
+   however, other options, namely RSH may also be available.
+
+   If using the SSH config to set your remote username ensure RSync is configured
+   to use the SSH transport layer, or that the chosen transport layer is itself
+   appropriately configured.
+
```
oliver-sanders, line 43:
```suggestion
```

src/7-to-8/major-changes/remote-owner.rst
oliver-sanders:
```suggestion
   * If you set task's ``[runtime][remote]owner`` configuration.
```

src/7-to-8/major-changes/remote-owner.rst
oliver-sanders: Makes sense, cheers.

hjoliver (approved):

wxtim (approved):
Content which you've done seems ok, and builds.

Do you want to have a signpost to this in the `summary.rst` - just a header and a link so that people who just read that get signposted (not going to stop PR on that basis)?

oliver-sanders:
Added a bullet point in the summary.

PR #486, opened 2022-06-28
Update platforms admin doc.

hjoliver:
Sibling of https://github.com/cylc/cylc-flow/pull/4938

**Requirements check-list**

### DIFF
--- a/src/reference/config/writing-platform-configs.rst
+++ b/src/reference/config/writing-platform-configs.rst
@@ -145,20 +145,25 @@ Background Jobs on Cluster with Other Options
    - Allow some background jobs to use an alternative shell,
      or an alternative ssh command.
 
+.. note::
+
+   The "background" and "at" job runners require single-host platforms,
+   because the job ID is only valid on the submission host.
+
 .. code-block:: cylc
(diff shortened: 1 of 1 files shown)


src/reference/config/writing-platform-configs.rst
```diff
@@ -167,7 +172,7 @@ Background Jobs on Cluster with Other Options
 Submit PBS Jobs from Localhost
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-- **Platforms can share hosts and not share batch systems.**
+- **Platforms can share hosts without sharing batch systems.**
```
hjoliver, line 175: (Off topic, but as-is this could be interpreted as saying "platforms cannot share batch systems"?)

wxtim, line 175: yes, but ...

```suggestion
- **Platforms can share hosts without sharing job runners.**
```

wxtim (approved):
I think this needs only one review and can be merged once it's primary is merged.

wxtim (approved):

PR #490, opened 2022-07-04
Improve migration guide

MetRonnie:
Add content to and reorganise into more logical order:
- Back compat
- Optional outputs/suicide triggers
- Platforms

Also update slightly incorrect `CYLC_VERSION` Jinja2 variable thing

Would really like to merge this by Wed 6th morning (UTC) + redeploy 8.0rc3 docs in time for workshop, sorry for the short notice!

**Requirements check-list**


hjoliver:
(I'll get onto this review pronto...)

src/7-to-8/major-changes/cli.rst
```diff
@@ -42,6 +42,9 @@ renamed or removed.
   - We no longer include a command for locating this documentation.
```
hjoliver, line 42: Commands like `cylc documentation` and `cylc checkpoint` that have been removed should say "removed" as the first sub-bullet point. (The intro says "changed or removed" but then the reader has to figure out which is which).

src/7-to-8/major-changes/cli.rst
```diff
@@ -42,6 +42,9 @@ renamed or removed.
...
 ``cylc edit``
   - Use a text editor to edit the workflow configuration file.
+``cylc hold``
+  - Now used on tasks only; use ``cylc pause`` to pause an entire workflow
+    (halt all job submissions).
```
hjoliver, line 47:
```suggestion
    (i.e. to halt all job submissions).
```

src/7-to-8/major-changes/cli.rst
```diff
@@ -105,8 +108,8 @@ In Cylc 7 there were two ways to specify a task:
    task.cycle
    cycle/task
 
-At Cylc 8 we have removed the former and extended the latter to provide a
-unique identifier for all workflows, cycles, tasks and jobs using a
+In Cylc 8 the former is now deprecated, and the latter has been extended to
+provide a unique identifier for all workflows, cycles, tasks and jobs using a
 standardised format:
```
hjoliver, line 113:
```suggestion
standard format:
```
(Not sure it has been "standardised" as such)

src/7-to-8/major-changes/compatibility-mode.rst
```diff
@@ -39,22 +40,101 @@ The old ``suite.rc`` filename triggers a backward compatibility mode in which:
...
+
+Python 2 to 3
+^^^^^^^^^^^^^
+
+Whereas Cylc 7 runs using Python 2, Cylc 8 runs using Python 3. This affects:
+- modules imported in Jinja2
+- Jinja2 filters, tests and globals
+- custom xtrigger functions
+
+Note that task scripts are not affected - they can run using an independent
```
hjoliver, line 84:
```suggestion
Note that task scripts are not affected - they run in an independent
```
(Not necessarily 100% true for local background jobs, but users should think of it like this anyway).

src/7-to-8/major-changes/compatibility-mode.rst
```diff
@@ -39,22 +40,101 @@ The old ``suite.rc`` filename triggers a backward compatibility mode in which:
...
+environment.
+
+See :ref:`py23` for more information and examples of how to implement
+interoperability.
```
hjoliver, line 88:
```suggestion
See :ref:`py23` for more information and examples of how to implement
interoperability if your workflows extend Cylc or Jinja2 with custom Python scripts.
```
(Should make it clear that most users do not need to worry about this).

src/7-to-8/major-changes/compatibility-mode.rst
```diff
@@ -106,149 +183,91 @@ You must change the configuration yourself. In this case:
...
-               batch system = pbs
+       [[foo]]
+           {% if CYLC_VERSION is defined and CYLC_VERSION[0] == '8' %}
+               script = cylc pause "$CYLC_WORKFLOW_ID"
+           {% else %}
+               script = cylc hold "$CYLC_SUITE_NAME"
+           {% endif %}
 
+Note this logic (and the ``CYLC_VERSION`` Jinja2 variable) is executed locally
+prior to parsing the workflow configuration.
```
hjoliver, line 237:
```suggestion
prior to Cylc parsing the workflow configuration.
```
(since Jinja2 does parse it earlier)

hjoliver (commented):
Partially reviewed; looks good so far.

src/7-to-8/major-changes/cli.rst
MetRonnie: I've split them into 3 lists:
- commands that have been removed entirely
- those which have been replaced
- those which have changed

src/7-to-8/major-changes/compatibility-mode.rst
```diff
@@ -39,22 +40,101 @@ The old ``suite.rc`` filename triggers a backward compatibility mode in which:
...
+
+Custom remote installation
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If you have used Rose 2019 you may be used to all files and directories inside
+your :term:`source directory` being installed on remote install targets.
+However, by default, Cylc 8 only installs
+:ref:`certain files and directories <RemoteInit>` onto remote install targets.
+See :ref:`below <compat.eg.custom_remote_install>` for an example of how to specify
+custom files to be installed.
```
oliver-sanders, line 115: This should go in / be covered by the install section - https://cylc.github.io/cylc-doc/latest/html/7-to-8/major-changes/cylc-install.html#remote-installation

src/7-to-8/major-changes/compatibility-mode.rst
```diff
@@ -106,149 +183,91 @@ You must change the configuration yourself. In this case:
...
-   Cylc 6 syntax deprecated by Cylc 7 is now obsolete. Run ``cylc validate``
-   *with Cylc 7* on your ``suite.rc`` to check for deprecation warnings and fix
-   those before validating with Cylc 8.
+.. code-block:: cylc
 
-.. warning::
+   {% if CYLC_VERSION is defined and CYLC_VERSION[0] == '8' %}
+   [scheduler]
+       install = my-dir/, my-file
+   {% endif %}
```
oliver-sanders, line 254: I get the 7/8 interop angle because of the config changed but I think this would be better off in the installation section for a cleaner separation rather than dipping into install without explaining it fully here, then covering it again there.

src/7-to-8/major-changes/suicide-triggers.rst
```diff
@@ -1,64 +1,57 @@
...
-Cylc 8 handles :term:`graphs <graph>` in an event-driven manner which means
-that a workflow can follow different paths in different eventualities without
-the need for suicide triggers. This is called :term:`graph branching`.
+   Typically this will be the case if you are using any of the following in
+   the workflow graph:
 
-.. seealso::
+   - :term:`suicide triggers <suicide trigger>`
+   - the ``|`` (OR) operator
+   - the ``:fail`` output or custom ouputs
```
oliver-sanders, line 27: These two are really just consequences of the former:

* Use of the OR operator does not mean there is graph branching.
  * e.g. `a | b => c` means `c` runs when the first of `a` or `b` succeed, it does not necessarily infer branching.
* At Cylc 7 use of the `:fail` output must always be accompanied by a suicide trigger.
   * Because failed tasks would not leave the pool unless forcefully ejected.

So I think this boils down to just suicide triggers?

Trying to come up with an example which doesn't?

wxtim (approved):

src/7-to-8/major-changes/suicide-triggers.rst
oliver-sanders: I've only come up with one example so far:

```
x:succeed | x:fail => y
```

This is equivalent to:

```
x:finish => y
```

Which is still valid at Cylc 8.

oliver-sanders: Ah yes, it's those psudo-outputs which cause the issue:

```
a:finish => b
a => c  # ERROR
```

This will also apply to the `:finish-any` and `:finish-all` family triggers.

src/7-to-8/major-changes/suicide-triggers.rst
```diff
@@ -1,64 +1,57 @@
-.. _728.suicide_triggers:
+.. TODO: rename to "Graph branching, optional outputs and suicide triggers", and give example (the one I posted on Element)
```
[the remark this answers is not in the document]
MetRonnie: Sorry, leftover from an earlier plan, will delete

src/7-to-8/major-changes/suicide-triggers.rst
```diff
@@ -1,64 +1,57 @@
...
-   # Remove the "bar" task in the fail case.
-   recover => ! bar
-
-   # Downstream dependencies.
-   bar | recover => baz
-
-Which results in the following logic:
+   foo => ! recover
+   # Remove the "foo" task in the fail case.
+   recover => ! foo
```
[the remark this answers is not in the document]
MetRonnie: I have taken out the first task because it was unnecessary for the example.

That first edge was confusingly commented as the success case, but it wasn't really, the `bar => baz` was the success case

```mermaid
graph TD
    subgraph before
    A(foo) -->|unnecessary| B(bar) -->|success case| C(baz)
    B -.->|fail case| D(recover) -.-> C
    end
```
```mermaid
graph TD
    subgraph after
    A(foo) --> B(bar)
    A -.-> C(recover) -.->B
    end
```

What part of it would you say was clearer before?

src/7-to-8/major-changes/suicide-triggers.rst
MetRonnie: I'll just mention suicide triggers then.

I was playing around with `:finish` and found a possible bug in stall behaviour for back-compat mode: https://github.com/cylc/cylc-flow/issues/4965

oliver-sanders: Ooh, good catch, that's nasty.

oliver-sanders: From a quick grep there are quite a lot of `:finish` triggers out there, to be expected. I think this is the only non-suicide case (?) but it is a biggie!

hjoliver: > At Cylc 7 use of the :fail output must always be accompanied by a suicide trigger.
> -  Because failed tasks would not leave the pool unless forcefully ejected.

I wouldn't say that's true - manual removal of failed tasks, if they don't need to be retriggered, is pretty common.

hjoliver:
> I have taken the (squashed) commits from this branch as-is and added to [branch-8.0rc3](https://github.com/cylc/cylc-doc/tree/branch-8.0rc3) so I can re-deploy the docs. I will do that now, can always re-deploy again if any further changes are needed

Ah, that explains why I was seeing your changes here, online already!! :confused:

src/7-to-8/major-changes/suicide-triggers.rst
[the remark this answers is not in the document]
hjoliver: Yeah the first task was unnecessary. It was probably abstracted from an old example that was meant to look more realistic.

src/7-to-8/major-changes/compatibility-mode.rst
```diff
@@ -39,22 +40,101 @@ The old ``suite.rc`` filename triggers a backward compatibility mode in which:
...
 
-If tasks in your workflow call Cylc commands directly it might be necessary to
-modify them to be compatible with command line interface changes
-(unfortunately, this is not possible to detect at validation).
+.. note::
+
+   Any previous-cycle workflow data needed by the new run will need to be
+   manually copied over from the original run directory.
+
+
```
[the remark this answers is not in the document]
hjoliver: It will be a very common requirement though. Should we not provide advice on how to do it? 

Maybe don't call it "restarting", to avoid confusion.  Perhaps something like **How to start your migrated workflow from where Cylc 7 left it**. And give general advice and caveats.

src/7-to-8/major-changes/suicide-triggers.rst
hjoliver: > Ooh, good catch, that's nasty.

Nasty bug fixed, phew.

src/7-to-8/major-changes/cli.rst
```diff
@@ -27,61 +27,79 @@ renamed or removed.
...
+``cylc reset``
+  - It is no longer possible to manually change a task's state.
+  - You can, however, override the outputs the task generated which has a
+    similar effect with ``cylc set-outputs``.
+``cylc restart``
+  - Replaced by ``cylc play``.
+``cylc run``
+  - Replaced by ``cylc play``.
+``cylc spawn``
+  - Spawning is now performed automatically, use ``cylc trigger`` to run a task.
```
hjoliver, line 91:
```suggestion
  - Spawning is now performed automatically, on demand. Use ``cylc trigger`` to run
    a task, or ``cylc set-outputs`` to spawn tasks that depend on specified outputs.
```

src/7-to-8/major-changes/cli.rst
```diff
@@ -27,61 +27,79 @@ renamed or removed.
...
+``cylc suite-state``
+  - Renamed as ``cylc workflow-state``.
```
hjoliver, line 93:
```suggestion
  - Renamed to ``cylc workflow-state``.
```

src/7-to-8/major-changes/cli.rst
```diff
@@ -27,61 +27,79 @@ renamed or removed.
...
+
+.. rubric:: Commands that have changed:
+
+``cylc hold``
+  - Now used on tasks only; use ``cylc pause`` to pause an entire workflow
+    (i.e. to halt all job submissions).
+``cylc release``
+  - Now used on tasks only; use ``cylc play`` to resume a paused workflow.
```
hjoliver, line 101:
```suggestion
  - Now used only to release held tasks; use ``cylc play`` to resume a paused workflow.
```

src/7-to-8/major-changes/compatibility-mode.rst
```diff
@@ -6,13 +6,14 @@ Cylc 7 Compatibility Mode
 .. admonition:: Does This Change Affect Me?
    :class: tip
 
-   This will affect you if you want to run run Cylc 7 workflows (using the
-   ``suite.rc`` filename) using Cylc 8.0
+   This will affect you if you want to run Cylc 7 (``suite.rc``) workflows
+   using Cylc 8.0
```
hjoliver, line 10:
```suggestion
   using Cylc 8
```
(Avoid having to update at 8.1)

hjoliver (approved):
LGTM!

src/7-to-8/major-changes/compatibility-mode.rst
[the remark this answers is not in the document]
MetRonnie: Or `"Restarting" a Cylc 7 workflow`?

src/7-to-8/major-changes/cli.rst
MetRonnie: ["Renamed as" seems to be more favoured](https://books.google.com/ngrams/graph?content=renamed+as%2Crenamed+to&year_start=1800&year_end=2019&corpus=26&smoothing=3&direct_url=t1%3B%2Crenamed%20as%3B%2Cc0%3B.t1%3B%2Crenamed%20to%3B%2Cc0)

hjoliver: Actually, I agree, on reflection.

src/7-to-8/major-changes/compatibility-mode.rst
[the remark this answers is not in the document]
oliver-sanders: > It will be a very common requirement though.

I'm not aware of any requirement or desire at our end. I think the only use case would be really, really, long-running workflows, I'm not aware that we have any examples of this. If they exist I would advise finishing the workflow run with Cylc 7 rather than to attempting to migrate to Cylc 8 part-way through a run.

Migrating like this is hazardous, e.g. remote installation works differently, data might need to be relocated on multiple filesystems, etc. We discussed this in a meeting at our end?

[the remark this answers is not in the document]
MetRonnie: I've replaced this subsection with a bullet point saying you can't restart, use Cylc 7 to finish a Cylc 7 workflow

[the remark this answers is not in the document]
hjoliver: > I'm not aware of any requirement or desire at our end. I think the only use case would be really, really, long-running workflows, I'm not aware that we have any examples of this.

Most of our workflows are "really, really, long-running" - i.e. they no final cycle point, or a very distant one. And most are "warm cycling" - i.e there are inter-cycle dependencies, and doing a cold start from scratch is not something you want to do unless you have no choice.

Anyhow, we may have got our wires crossed here. I just meant "advise users how to continue a warm-cycling Cylc 7 workflow with Cylc 8", not a literal in-place restart from the DB. 

Along the lines of:

*Shut down (with Cylc 7) after a particular cycle point, copy any inter-cycle data over to a new Cylc 8 install (taking account of the run-name if appropriate), and cold start that with Cylc 8 at the next cycle point.*

src/7-to-8/major-changes/suicide-triggers.rst
```diff
@@ -71,35 +59,52 @@ Which results in the following logic:
...
+
+.. code-block:: console
+
+   $ cylc validate .
+   GraphParseError: Opposite outputs foo:succeeded and foo:failed must both be optional if both are used
 
-In Cylc 8 we mark the :term:`outputs <task output>` which are
-:term:`optional <optional output>` (in this case ``bar:success`` and
-``bar:fail``) with a ``?`` in the graph.
+In Cylc 8, all task outputs are :term:`expected <expected output>` to complete
```
hjoliver, line 73: NB #492 `s/expected/required/g`

hjoliver:
Two approvals, and time is of the essence this week. I'll post a small follow-up.  Thanks @MetRonnie

src/7-to-8/major-changes/suicide-triggers.rst
hjoliver: (I'll update 492 after merging this)

PR #497, opened 2022-07-18
Document how to continue a Cylc 7 run with Cylc 8.

hjoliver:
Follow-up https://github.com/cylc/cylc-doc/pull/490#discussion_r922915976 

Many Cylc 7 workflows (at NIWA anyway!) "warm cycle" indefinitely, and cold-starting them is a last resort. 

E.g. our warm-cycled limited area atmos workflows can be cold started from global model data if absolutely necessary, but then forecast quality suffers over a subsequent spin-up period. 

Consequently, we can't just say "complete your run with Cylc 7". The best we can do is warn that that's preferred option *if possible*, and explain how to continue a run with Cylc 8, in a new run directory, if needed.

**Requirements check-list**

### DIFF
--- a/src/7-to-8/major-changes/continuing-c7-c8.rst
+++ b/src/7-to-8/major-changes/continuing-c7-c8.rst
@@ -8,8 +8,8 @@ Continuing a Cylc 7 Workflow with Cylc 8
 
    Read this if you have a partially complete Cylc 7 workflow that you want to
    continue, rather than start from scratch, with Cylc 8. Some cycling
-   workflows, for example, may need to run expensive "cold start" tasks
-   and incur a multi-cycle spin-up if started from scratch.
+   workflows, for example, may need to run expensive "cold start" tasks and
+   incur a multi-cycle spin-up if started from scratch.
 
 .. warning::
 
@@ -35,7 +35,12 @@ To continue a Cylc 7 workflow with Cylc 8:
 
    - This could include external files installed by initial tasks at runtime
    - Note different files could be present on different job platforms
-4. Start the new Cylc 8 run at the appropriate cycle point or task(s)
+4. Start the new Cylc 8 run at the appropriate cycle point or task(s) in the
+   graph
 
-   - To avoid re-running initial tasks, don't reset the :term:`initial
-     cycle point` to the new :term:`start point <start cycle point>`
+   - Don't reset the :term:`initial cycle point` (in the ``flow.cylc`` or on
+     the command line) to the :term:`start point <start cycle point>` of the
+     Cylc 8 run. That would result in the "cold start" that this continuation
+     procedure is designed to avoid. Instead use the ``--start-cycle-point``
+     option (or ``--start-task``) with ``cylc play``, to start at the right
+     place within the graph.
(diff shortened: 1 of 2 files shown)


src/7-to-8/major-changes/continuing-c7-c8.rst
```diff
@@ -0,0 +1,41 @@
...
+     when a job is first submitted to a platform
+3. Copy runtime files needed by upcoming tasks from the old to the new run
+   directory
+
+   - This could include external files installed by initial tasks at runtime
+   - Note different files could be present on different job platforms
+4. Start the new Cylc 8 run at the appropriate cycle point or task(s)
+
+   - To avoid re-running initial tasks, don't reset the :term:`initial
+     cycle point` to the new :term:`start point <start cycle point>`
```
MetRonnie, line 41: I'm not sure what the bullet point means. Perhaps it should just say "use the `--startcp`" option?

MetRonnie (approved):

src/7-to-8/major-changes/continuing-c7-c8.rst
hjoliver: I mean, don't change `initial cycle point` (in the `flow.cylc` or via the command line) to the start cycle of the Cylc 8 run, because that would result in a new cold start (i.e., from the beginning of the graph, with any attendant special behaviour in early cycles) that we're presumably trying to avoid with this set up procedure.

I'll expand on that a bit...

MetRonnie (approved):

src/7-to-8/major-changes/continuing-c7-c8.rst
```diff
@@ -0,0 +1,46 @@
...
+   - Typically the end of a cycle point, to simplify the continuation
+2. :ref:`Install <MajorChangesInstall>` a new instance of the workflow from
+   source, with Cylc 8
+
+   - Adapt file paths to the new run directory structure, in workflow and task
+     configurations
+   - Note Cylc 8 does :ref:`remote file installation <728.remote-install>`
+     when a job is first submitted to a platform
+3. Copy runtime files needed by upcoming tasks from the old to the new run
+   directory
```
oliver-sanders, line 34: This is ok for local hosts, however, on remote hosts it's trickier due to remote installation.

I think ATM we allow remote-installation to install "over the top of" an existing run dir? However, there's an issue open to change that which would require this step to occur *after* remote installation.

Maybe mention `work/` and `share/`?

hjoliver: I've made two fairly open-ended comments that are relevant:
> - Note Cylc 8 does :ref:`remote file installation <728.remote-install>`
     when a job is first submitted to a platform
> - Note different files could be present on different job platforms

But OK, I'll be a bit more explicit, on those points. (Later...)

hjoliver: (Just saw I haven't addressed this, and I've run out of time today ... but the existing text is not actually wrong, so I can do a quick follow-up PR later if this gets merged first).

dwsutherland (approved):
Makes sense to me.. built fine;
image

In Cylc7 I would often just start at a new CP and set cold-start to succeeded.. This would avoid that, unless you want to run some (but not all) initial cycle stuff (i.e. deployment).

PR #501, opened 2022-07-21
Add Spellchecking to Cylc Doc

wxtim:
Another attempt to add a spelling check to Cylc Docs. 

If accepted there are questions to be answered about some of the words on the wordlist, but IMO these are follow-up issues.

**Requirements check-list**

# Follow up questions which should (IMO) be out of scope:

- Should we decide if we are using "parameterize" , "parameterise", "parametrize" (consistent with Pytest) or "parametrise" - As far as I can tell they are all valid (not paying a subscription but all seem to be in use and in the OED.) Multiple spelling are common, which is ugly - my vote is "parametrize" because then it's consistent with pytest.
- Should we allow lower-case job-runner names?
- Do we want to take some words out of the list ("schd"?) and put them in documents with the spelling directive.


src/dictionaries/proper_nouns
```diff
@@ -0,0 +1,40 @@
+Aldergrove
+Boolean
+Booleans
+Camborne
+conda
+Conda
```
oliver-sanders, line 6: ?

src/dictionaries/words
```diff
@@ -0,0 +1,214 @@
+
+
+
+
+
+
+
+
+
```
oliver-sanders, line 9: ?

spellcheck
```diff
@@ -0,0 +1,398 @@
+diff --git a/.github/workflows/test.yml b/.github/workflows/test.yml
```
oliver-sanders, line 1: ?

oliver-sanders (commented):
Willing to give it a try.

Note, I don't know if it will catch autodocumented spellings.

spellcheck
wxtim: Something has gone wrong. Will investigate.

src/dictionaries/proper_nouns
wxtim: What are you questioning?

> Why both upper and lower case (programmatic)?

Conda gets used as a command as well as a proper noun. Without more sophisticated context we need to allow for both.

> Why both upper and lower case (grammatical)?

You need both, because I'm not attempting to auto generate the sentence case versions of words in the "proper_nouns" list. You have to explicitly state proper nouns you want to allow a lower case version of.

src/dictionaries/words
wxtim: No. Will examine. Something not OK.

wxtim:
> Note, I don't know if it will catch autodocumented spellings.

I think that it does.

src/dictionaries/words
wxtim: Fixed

src/conf.py
```diff
@@ -245,3 +247,33 @@
...
+
+# Sort wordlist:
+words = wordsfile.read_text().split('\n')
+words = [w for w in sorted(words) if w]
+wordsfile.write_text('\n'.join(words) + '\n')
+
+# Create sentence case versions of wordlist:
+sentence_case = []
+for word in words:
+    sentence_case.append(word.capitalize())
```
MetRonnie, line 278:
```suggestion
sentence_case = [word.capitalize() for word in words]
```

MetRonnie (approved):

src/dictionaries/proper_nouns
oliver-sanders: > Conda gets used as a command as well as a proper noun. Without more sophisticated context we need to allow for both

The same applies to Cylc but we only have the capitalised entry:

```rst
To open Cylc TUI run the command ``cylc tui``.
```

Presumably Conda should be held to the same standard?

src/dictionaries/words
```diff
@@ -0,0 +1,205 @@
+app
+apps
+args
+auth
+authenticator
+bar
+baz
+behavior
+behaviour
```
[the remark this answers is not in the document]
hjoliver: Does the dictionary have locale variants?  Otherwise I imagine it would either allow British and American spellings, or enforce American?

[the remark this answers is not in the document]
MetRonnie: At the moment Tim has hardcoded it to `en_NZ` (I don't know what the differences to `en_GB` are)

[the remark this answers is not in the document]
wxtim: I've deliberately allowed both for now. 

I think we should come to a team decision and pick one as follow up - see comments on "Parametrize". 

Since it's not a keyword AFAIK in any languages we use (my reason for "color", used in HTML to use US spelling) I think "u" should be in there.

src/dictionaries/words
```diff
@@ -0,0 +1,205 @@
...
+colored
+coloured
+conf
+config
+configargs
+configs
+coroutine
+coroutines
+cyclepoint
+cylc
```
wxtim, line 21:
```suggestion
```

src/dictionaries/proper_nouns
```diff
@@ -0,0 +1,40 @@
...
+Cylc
```
wxtim, line 7:
```suggestion
Cylc
cylc
```

src/dictionaries/proper_nouns
wxtim: > The same applies to Cylc but we only have the capitalised entry:

Not true, I just put "cylc" in the word list by mistake.

MetRonnie (approved):

src/dictionaries/proper_nouns
```diff
@@ -0,0 +1,41 @@
+Aldergrove
+Boolean
+Booleans
```
hjoliver, line 3: So plurals aren't handled automatically?

src/dictionaries/proper_nouns
```diff
@@ -0,0 +1,41 @@
...
+Camborne
+conda
+Conda
+Cylc
+cylc
+Datapoint
+Empy
+EmPy
```
hjoliver, line 11: Is one of these not the correct one?

src/dictionaries/words
```diff
@@ -0,0 +1,204 @@
...
+auth
+authenticator
+bar
+baz
+behavior
+behaviour
+boolean
+booleans
+colored
+coloured
```
hjoliver, line 13: Like the other terms under discussion, I think we should decide one way or the other (US or not) and be consistent. It  looks a bit shonky having a random selection of multiple "correct" spelling.

hjoliver (approved):
A nice improvement :+1: 
Couple of minor questions, for the meeting.

src/dictionaries/proper_nouns
wxtim: No, and not surprising given the irregularity of English plurals.

src/dictionaries/proper_nouns
wxtim: The first, I think.

src/dictionaries/words
wxtim: It's horrid, but I propose putting it in like this and creating a fix issue, because I agree, it's not nice.

src/dictionaries/proper_nouns
MetRonnie: "EmPy" is the correct capitalisation

hjoliver:
Project meeting decision: UK/NZ spelling in the cylc-doc text.  US spelling for programming (or closely related) terms in config files etc. - which as string literals should transfer to the docs without special treatment.

src/dictionaries/words
hjoliver: That's fine with me (I guess we have to delete the new dictionary entries *and* fix every instance in the docs?)

hjoliver:
OK, merging. I'll post a follow-up issue for the other stuff.

PR #534, opened 2022-09-09
Install Target and Platform Update

datamel:
This adds more information about install targets and clarifies some of the platform documentation.  

**Requirements check-list**


src/7-to-8/major-changes/platforms.rst
```diff
@@ -210,13 +210,20 @@ And the platform settings for these examples might be:
...
            # This computer with Slurm requires you to use a login node.
            hosts = login_node01, login_node02  # Cylc will pick a host.
            job runner = slurm
 
 
+Note that in these examples, it is assumed that ``linuxboxNN``, ``pbs_local`` and
+``slurm_supercomputer`` have distinct file systems.
+If platforms share a file system they must have their :term:`install target`
+configured, this should be done at site level. For more information see
+:ref:`Install Targets`.
```
wxtim, line 225:
```suggestion
If an install target is not set, a platform will use its own platform name
as the install target name. If multiple platforms share a file system
and :ref:`install target <Install Targets>` is not set each will carry out a separate (slow) remote
install procedure. Sets of platforms which share a file system should specify
a single :ref:`install target <Install Targets>`.
```

src/reference/config/writing-platform-configs.rst
```diff
@@ -60,37 +60,94 @@ What Are Install Targets?
...
+Check the second platform for the file:
+
+   .. code-block:: console
+
+      $ ssh platform-B
+      $ cat cylc-test-file
+      $ exit
+
+If the test file exists, then these two platforms share a file system and will
+require the same install target in ``global.cylc`` config file.
```
wxtim, line 92:
```suggestion
If the test file exists, then these two platforms share a file system and should
use the same install target in ``global.cylc`` config file.
```

src/reference/config/writing-platform-configs.rst
```diff
@@ -60,37 +60,94 @@ What Are Install Targets?
...
+
+Example Platform Configurations
+-------------------------------
 
 .. admonition:: Scenario
 
-   You want to allow users to submit small jobs to the scheduler host:
+   Cylc scheduler hosts share a file system with a compute cluster.
```
wxtim, line 99: This is almost the same scenario as "Install PBS jobs from localhost"

src/reference/config/writing-platform-configs.rst
```diff
@@ -60,37 +60,94 @@ What Are Install Targets?
...
+.. code-block:: cylc
+   :caption: the ``global.cylc`` config file for this scenario could look like:
+
+   [platforms]
+       [[cluster_1]] # cylc-scheduler platform
+           hosts = host_1, host_2
+           # unset install target defaults to platform name: cluster_1
+       [[cluster_2]]
+           hosts = host_3, host_4
+           # unset install target defaults to platform name: cluster_2
```
wxtim, line 145: I find the word "unset" confusing:
```suggestion
   [platforms]
       [[cluster_1]] # cylc-scheduler platform
           hosts = host_1, host_2
           # install target defaults to platform name: cluster_1
       [[cluster_2]]
           hosts = host_3, host_4
           # install target defaults to platform name: cluster_2
```

or even
```suggestion
   [platforms]
       [[cluster_1]] # cylc-scheduler platform
           hosts = host_1, host_2
           # install target = platform name (cluster_1)
       [[cluster_2]]
           hosts = host_3, host_4
           # install target = platform name (cluster_2)
```

src/reference/config/writing-platform-configs.rst
```diff
@@ -103,7 +160,7 @@ Simple Remote Platform
    [platforms]
        [[myhost]]
            hosts = myhost
-           install target = myhost
+           # unset install target defaults to myhost
```
wxtim, line 163: Not sure I like the unset word.

wxtim (approved):
IMO a worthwhile change even if you bin my suggestions - so approved.

src/7-to-8/major-changes/platforms.rst
datamel: If they share a file system the authentication keys will clash (resulting in remote init error) if they have different install targets configured though?

wxtim: Ok, then you are right, they must. I still think that the clarification makes sense

```suggestion
Sets of platforms which share a file system must specify
a single :ref:`install target <Install Targets>`.

.. note::
   If an install target is not set, a platform will use its own platform name
   as the install target name. If multiple platforms share a file system
   and :ref:`install target <Install Targets>` they will not work.
```

src/7-to-8/major-changes/platforms.rst
```diff
@@ -72,7 +72,7 @@ Platforms also define how hosts are selected from each platform:
 
 There may be cases where sets of platforms (for example a group of
 standalone compute servers, or a pair of mirrored HPC's) might be equally
-suitable for a task, but not share files systems to allow them to constitute
+suitable for a task, but not share file systems to allow them to constitute
```
dpmatthews, line 75: Whether they share file systems or not isn't really relevant.
I suggest ending the sentence after "equally suitable for a task".

src/7-to-8/major-changes/platforms.rst
```diff
@@ -210,13 +210,20 @@ And the platform settings for these examples might be:
            # A computer with PBS, that takes local job submissions
            job runner = pbs
            hosts = localhost
+           install target = localhost
 
        [[slurm_supercomputer]]
```
dpmatthews, line 215: The example above uses a platform named "just_run_it" - it should be "slurm_supercomputer".

datamel:
Converting to draft after discussion with @dpmatthews. Examples will be simplified to three main use cases.

MetRonnie:
Is this an 8.0.x addition? If so need to change the merge base to the 8.0.x branch

src/7-to-8/major-changes/cylc-install.rst
```diff
@@ -89,7 +89,7 @@ Remote Installation
 
 Remote file installation does not occur until running the workflow.
 When the first task runs on a remote platform, Cylc will transfer files from
-the :term:`run directory` to the install target.
+the :term:`run directory` to the :term:`install target <install target>`.
```
dpmatthews, line 92: Why the difference in syntax for `run directory` vs `install target`?

datamel: > Why the difference in syntax for `run directory` vs `install target`?
https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#cross-referencing-syntax

The `<>` is for linking to a title, different to what is displayed. In this case it is redundant as install target is the same, typed out of habit.

src/installation.rst
```diff
@@ -240,7 +240,7 @@ Configuration
 Cylc uses "sane and safe" defaults and is suitable for use "out of the box".
```
dpmatthews, line 240: Maybe add "... if all you need to do is run jobs locally in the background"?

src/installation.rst
```diff
@@ -240,7 +240,7 @@ Configuration
...
 However, many things may need to be configured, e.g:
 
-* Job hosts
+* Job hosts, see :ref:`AdminGuide.PlatformConfigs`
 * Communication methods
 * User/Site preferences
```
dpmatthews, line 245: Comms methods are part of platforms and user/site preferences is a bit generic.
I think there are other sections it would be helpful to link to.
How about (syntax not checked):
```
* :ref:`AdminGuide.PlatformConfigs` (jobs hosts, runners, etc)
* :ref `Submitting Workflows To a Pool Of Hosts <Scheduler hosts>`
* :ref `Event Handling <Default event handlers>`
```

src/reference/config/writing-platform-configs.rst
```diff
@@ -49,7 +49,7 @@ Why Were Platforms Introduced?
 
 - Allow a compute cluster with multiple login nodes to be treated as a single
   unit.
-- Allow Cylc to elegantly handle failure of to communicate with login nodes.
+- Allow Cylc to elegantly handle failure to communicate with login nodes.
 - Configure multiple platforms with the same hosts; for example you can use
   separate platforms to submit jobs to a batch system and to background on
   ``localhost``.
```
dpmatthews, line 55: I don't think this a valid reason why platforms were introduced - this was just as easy before.
Instead we could add:
```
- Reduce the number of ssh connections required for job submission and polling.
```

src/reference/config/writing-platform-configs.rst
```diff
@@ -60,114 +60,195 @@ What Are Install Targets?
...
-compute cluster. Cylc does not need to install files on this cluster. The
-cluster and scheduler hosts are different platforms, but share an install
-target.
+Note, if missing from configuration, the install target will default to the
+platform name. If incorrectly configured, this will cause errors in
+:ref:`RemoteInit`.
 
-But you might also have mirrored clusters, each with their own file system.
-Each cluster would be a platform with its own install target.
+To check if two platforms share a file system:
```
dpmatthews, line 73: I would remove this bit - I don't think we need to explain how to do this

src/reference/config/writing-platform-configs.rst
```diff
@@ -60,114 +60,195 @@ What Are Install Targets?
...
+-------------------------------
 
-- **Platforms don't need to be complicated**
-- ``install target`` **specifies a file system for the task using that platform**
+Detailed below are some examples of common platform configurations.
+
+Multiple Platforms Sharing File System with Cylc Scheduler
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+- **The Scheduler Host (Cylc Server) provides a built in localhost platform**
```
dpmatthews, line 102: Suggest replacing this with:
```
The :term:`scheduler` runs on the ``localhost`` platform.
```
and add a link to the new section added in #542

src/reference/config/writing-platform-configs.rst
```diff
@@ -60,114 +60,195 @@ What Are Install Targets?
...
-   :caption: part of a ``global.cylc`` config file
+   :caption: the ``global.cylc`` config file for this scenario could look like:
 
    [platforms]
-       [[myhost]]
-           hosts = myhost
-           install target = myhost
+       [[localhost]]
+           hosts = localhost
+           install target = localhost
```
dpmatthews, line 125: These lines are not needed and should be removed.

src/reference/config/writing-platform-configs.rst
```diff
@@ -60,114 +60,195 @@ What Are Install Targets?
...
+       [[desktop\d\d\d]]
+           install target = localhost
 
+The ``localhost`` platform is the Cylc Scheduler host, as configured in
+:cylc:conf:`global.cylc[scheduler][run hosts]available`. This is the host that
+the workflow will start on.
```
dpmatthews, line 131: Remove these lines - now covered above?

src/reference/config/writing-platform-configs.rst
dpmatthews: Ah, I see that they get removed in the text below.
I think it might be simpler to just explain that the localhost  platform does not need to be configured in this case since it is setup by default.

src/reference/config/writing-platform-configs.rst
```diff
@@ -60,114 +60,195 @@ What Are Install Targets?
...
+The ``desktop\d\d\d`` platform will set up 1000 platforms, all with the same
+specification and one host per platform.
```
dpmatthews, line 133: Suggest:
```
 ``desktop\d\d\d`` is a pattern which defines multiple platforms.
When using a pattern the "hosts" setting must be left unset so that it defaults to the platform name.
This ensures each of the matching platforms is unique.
```

src/reference/config/writing-platform-configs.rst
```diff
@@ -60,114 +60,195 @@ What Are Install Targets?
...
+
+.. note::
+
+   Cylc carries out a "fullmatch" regular expression comparison with the
+   the platform name so ``desktop\d\d\d`` is effectively the same as
+   ``^desktop\d\d\d$``.
+
+With this config setup, Cylc is aware that these two platforms do not require
+remote installations. Since the job runner is not set, this will default to
```
dpmatthews, line 142: Suggest:
```
With this config setup, Cylc is aware that the ``desktop\d\d\d`` platforms share
the same file systems as the scheduler host so do not require remote installation.
```

src/reference/config/writing-platform-configs.rst
```diff
@@ -60,114 +60,195 @@ What Are Install Targets?
...
+
+      [runtime]
+          [[mytask]]
+              platform = desktop123
+
+in their workflow configuration.
+If ``flow.cylc[runtime][mytask]platform`` is unset, the job will run on the Cylc
+Scheduler host using this default ``localhost`` platform. It may be appropriate
+to allow users to run small jobs on the Cylc Server however more intensive jobs
+should run on a more appropriate platform, as the next scenario will detail.
```
dpmatthews, line 176: Cylc can be used to run workflows on large servers with all jobs running locally.
Suggest removing this last sentence.

src/reference/config/writing-platform-configs.rst
```diff
@@ -60,114 +60,195 @@ What Are Install Targets?
...
 
    You have a cluster where users submit to a single Slurm job queue from
-   either of a pair of identical login nodes which share a file system:
+   either of a pair of identical login nodes which share a file system.
 
 .. code-block:: cylc
    :caption: part of a ``global.cylc`` config file
 
    [platforms]
+       [[localhost]]  # Cylc Scheduler
```
dpmatthews, line 193: No need for this line?

src/reference/config/writing-platform-configs.rst
```diff
@@ -60,114 +60,195 @@ What Are Install Targets?
...
        [[spice_cluster]]
```
dpmatthews, line 194: "slurm_cluster" would be a more generic name!

MetRonnie:
As this appears to be compatible with 8.0.x I have changed the PR base to the 8.0.x branch

src/reference/config/writing-platform-configs.rst
```diff
@@ -60,114 +60,195 @@ What Are Install Targets?
...
 
 Cluster with Multiple Login Nodes
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+- **Platforms with multiple hosts require job runner to be set**
 - **Platforms can group multiple hosts together.**
 
 .. admonition:: Scenario
 
    You have a cluster where users submit to a single Slurm job queue from
```
dpmatthews, line 186: "to a single Slurm job queue" -> "jobs to Slurm"
(Slurm can have multiple queues)

src/reference/config/writing-platform-configs.rst
dpmatthews: (comment applies to the next line)

src/reference/config/writing-platform-configs.rst
```diff
@@ -49,277 +49,276 @@ Why Were Platforms Introduced?
...
 .. code-block:: cylc
    :caption: part of a ``global.cylc`` config file
 
    [platforms]
-       [[myhost]]
-           hosts = myhost
-           install target = myhost
+       [[localhost]]
+           hosts = localhost
+           install target = localhost
```
dpmatthews, line 103: This isn't required. I assume you want to include it for clarity?
You could have it commented out? e.g.
```
       # The localhost platform is available by default
       #[[localhost]]
       #    hosts = localhost
       #    install target = localhost
```
Or just remove it and mention it in the paragraph above.

src/reference/config/writing-platform-configs.rst
```diff
@@ -49,277 +49,276 @@ Why Were Platforms Introduced?
...
-- ``install target`` **specifies a file system for the task using that platform**
-
-.. admonition:: Scenario
+   You have a cluster where you can submit jobs from the Cylc scheduler host
+   using PBS.
 
-   Users want to run background jobs on a single server,
-   which doesn't share a file system with the workflow host.
+The ``localhost`` platform is the Cylc Scheduler host, as configured in
+:cylc:conf:`global.cylc[scheduler][run hosts]available`. This is the host that
```
dpmatthews, line 91: Add link to the section added in #542

src/reference/config/writing-platform-configs.rst
```diff
@@ -49,277 +49,276 @@ Why Were Platforms Introduced?
...
+           hosts = localhost
+           install target = localhost
+       [[pbs_cluster]]
+           hosts = localhost
+           job runner = pbs
+           install target = localhost
 
+Our Cylc scheduler does not have a job runner defined. Any job submitted to
+this ``localhost`` platform will run as a background job. Users can now set 
+:cylc:conf:`flow.cylc[runtime][<namespace>]platform` = pbs_cluster to run pbs jobs.
```
dpmatthews, line 111: pbs_cluster needs quoting

src/reference/config/writing-platform-configs.rst
```diff
@@ -49,277 +49,276 @@ Why Were Platforms Introduced?
...
-           job runner = slurm
-           install target = spice_cluster
-           retrieve job logs = True
-
-If either host is unavailable Cylc will attempt to start and communicate with
-jobs via the other login node.
-
-Since the platform hosts do not share a file system with the scheduler
-host we need to ask Cylc to retrieve job logs.
+       [[localhost]]
```
dpmatthews: Remove this line?

datamel:
Requesting review @wxtim as there have been significant changes since your approval, following discussions with @dpmatthews.

src/reference/config/writing-platform-configs.rst
```diff
@@ -49,277 +49,276 @@ Why Were Platforms Introduced?
...
-   Having hosts in a platform means that Cylc can communicate with
-   jobs via any host at any time. Platform groups allow Cylc to
-   pick a platform when the job is started, but Cylc will not then
-   be able to communicate with that job via hosts on another
-   platform in the group.
-
-
-Preferred and Backup Hosts and Platforms
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+               platforms = slurm_cluster, slurm_cluster_bg
```
dpmatthews, line 250: I don't think it makes sense to have a group mixing slurm & background jobs.
The example I was thinking of is wanting to run a background job on either of the login nodes so the config would be:
```
[platforms]
    [[slurm_cluster, slurm_cluster_bg1, slurm_cluster_bg2]]  # settings that apply to all:
        install target = slurm_cluster
        retrieve job logs = True
    [[slurm_cluster]]
        hosts = login_node_1, login_node_2
        job runner = slurm
    [[slurm_cluster_bg1]]
        hosts = login_node_1
    [[slurm_cluster_bg2]]
        hosts = login_node_2
[platform groups]
    [[slurm_cluster_bg]]
        platforms = slurm_cluster_bg1, slurm_cluster_bg2
```

src/reference/config/writing-platform-configs.rst
```diff
@@ -49,277 +49,276 @@ Why Were Platforms Introduced?
...
 
-- **You can set how hosts are selected from platforms.**
-- **You can set how platforms are selected from groups.**
+Group platforms together using the configuration item
+:cylc:conf:`global.cylc[platform groups]`. In the above example, the platforms
+``slurm_cluster`` and ``slurm_cluster_bg`` both share a file system (
+install target = ``slurm_cluster``). We advise caution when grouping platforms
+with different install targets as users could encounter a scenario whereby
+files (installed by Cylc during the remote initialization process) are
+not available to them.
```
dpmatthews, line 258: That's not right (I hope!).
I guess the bit in brackets should be "(created by a previous task using the same platform group)".

src/reference/config/writing-platform-configs.rst
```diff
@@ -49,277 +49,276 @@ Why Were Platforms Introduced?
...
+                   ...
 
-   Cylc carries out a "fullmatch" regular expression comparison with the
-   the platform name so ``desktop\d\d\d`` is effectively the same as
-   ``^desktop\d\d\d$``.
+These ``localhost`` symlinks are created during the cylc install process.
+Symlinks for remote install targets are created during :ref:`RemoteInit` following
+``cylc play``.
 
+Symlinks
```
dpmatthews, line 319: Not sure why this line is here?

src/glossary.rst
```diff
@@ -1051,6 +1051,14 @@ Glossary
 
          * :term:`reinstall`
 
+   install target
+      Cylc uses install targets to determine which hosts share file systems,
+      these are expected to be managed at site level and configured in
```
wxtim, line 1056:
```suggestion
      Cylc uses install targets to determine which platforms share file systems.
      Install targets should normally be managed at site level: They are configured in

File: src/installation.rst [hunk 30]
```diff
@@ -237,12 +237,13 @@ To do so create a symbolic link to the wrapper, for each of these commands:
 Configuration
 -------------
 
-Cylc uses "sane and safe" defaults and is suitable for use "out of the box".
+Cylc uses "sane and safe" defaults and is suitable for use "out of the box",
```
wxtim, line 240: Mr self deprecating suggests that we have a bit of faith in our product. The inverted commas imply that we don't fully believe that these settings are sane and safe.

```suggestion
Cylc uses sane and safe defaults and is suitable for use "out of the box",
```

src/reference/config/writing-platform-configs.rst
```diff
@@ -49,277 +49,282 @@ Why Were Platforms Introduced?
...
+- Reduce the number of ssh connections required for job submission and polling.
 
 .. _Install Targets:
 
 What Are Install Targets?
 ^^^^^^^^^^^^^^^^^^^^^^^^^
 
 Install targets represent file systems. More than one platform can use the
-same file system. It defaults to the name of the platform.
+same file system. Cylc relies on the site setup file ``global.cylc`` to determine
```
wxtim, line 61: I think we use "config" or "configuration" elsewhere. IMO "setup" is better, but consistency is better still.
```suggestion
same file system. Cylc relies on the site configuration file ``global.cylc`` to determine
```

src/reference/config/writing-platform-configs.rst
```diff
@@ -49,277 +49,282 @@ Why Were Platforms Introduced?
...
+which platforms share install targets. Cylc will then use this information to
+make the correct installations on remote platforms, including installation of
+files, creation of :cylc:conf:`global.cylc[install][symlink dirs]` and
+authentication keys to enable secure communication between platforms.
```
wxtim, line 65: Hemmingway says this sentence has a very high reading age. 

```suggestion
which platforms share install targets.

Cylc will setup each remote install target once. During setup it will:

  - Install workflow files
  - Symlink directories
  - Copy authentication keys (to allow secure communication)
```

src/glossary.rst
datamel: Can we either replace the capital with lowercase or the colon with a full stop?

```suggestion
      Cylc uses install targets to determine which platforms share file systems.
      Install targets should normally be managed at site level. They are configured in
```

src/reference/config/writing-platform-configs.rst
```diff
@@ -49,277 +49,282 @@ Why Were Platforms Introduced?
...
+required files which are on the scheduler host will be accessible on the
+desktop. From Cylc's point of view, the desktop and scheduler hosts are
+considered different platforms but must share an :term:`install target`.
+Cylc needs to be told that these platforms share an install target and so we
+configure this using the designated configuration item:
+:cylc:conf:`global.cylc[platforms][<platform name>]install target`.
+Cylc has optional configuration ``[[[meta]]]`` to add a description of the
+platform, this may be helpful to use, we will add a platform description to our
+desktop platform.
+The following platform definition is simplified, taking advantage of defaults
```
wxtim, line 142: - I think it's worth saying that it's the platform config (not Cylc) which has the `meta` section, since both workflows and tasks also have a meta section.

- I think paragraph breaks for readability.

- I've tried to make it clearer that metadata is there to help users. (In the workflow and task it can have a functional purpose.)

```suggestion
:cylc:conf:`global.cylc[platforms][<platform name>]install target`.

:cylc:conf:`global.cylc[platforms][<platform name>] has optional configuration 
``[[[meta]]]`` which users can view with ``cylc config --platforms``. We will add
a description designed to help users in this example.

The following platform definition is simplified, taking advantage of defaults
```

src/reference/config/writing-platform-configs.rst
```diff
@@ -49,277 +49,282 @@ Why Were Platforms Introduced?
...
-   jobs via any host at any time. Platform groups allow Cylc to
-   pick a platform when the job is started, but Cylc will not then
-   be able to communicate with that job via hosts on another
-   platform in the group.
-
-
-Preferred and Backup Hosts and Platforms
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+           [[slurm_cluster_bg]]
+               platforms = slurm_cluster_bg1, slurm_cluster_bg2
```
dpmatthews, line 253: Wrong indentation

src/reference/config/writing-platform-configs.rst
```diff
@@ -49,277 +49,282 @@ Why Were Platforms Introduced?
...
+       [[slurm_cluster]]
+           hosts = login_node_1, login_node_2
+           job runner = slurm
+           retrieve job logs = True
 
-But ``host`` defaults to ``localhost`` so you can simplify
-the ``[[pbs_cluster]]`` definition.
+The ``slurm_cluster`` hosts do not share a file system with the scheduler,
+therefore ``slurm_cluster`` is a remote platform.
+As the ``install target`` setting for each platform has been omitted, this will
```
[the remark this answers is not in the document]
datamel: 👍 Changed.

src/reference/config/writing-platform-configs.rst
datamel: This is fixed now.

dpmatthews (approved):

wxtim (approved):

PR #548, opened 2022-10-05
GH Actions: upgrade to `ubuntu-latest`; use conda to install graphviz

### LINKED ISSUE #247: graphviz: upgrade to latest
Due to a [graphviz issue](https://gitlab.com/graphviz/graphviz/-/issues/1855) we pinned our actions to Ubuntu 18 in https://github.com/cylc/cylc-doc/pull/246.

Once the issue is resolved we should unpin and update.

MetRonnie:
`ubuntu-18.04` is now deprecated - https://github.com/actions/runner-images/issues/6002

Closes #247

**Requirements check-list**

### DIFF
--- a/.github/actions/install-dependencies/action.yml
+++ b/.github/actions/install-dependencies/action.yml
@@ -1,5 +1,5 @@
 name: Install dependencies
-description: apt-get dependencies
+description: Install dependencies and setup + activate conda env
 
 runs:
   using: composite
@@ -15,7 +15,6 @@ runs:
         sudo apt-get install -y \
           librsvg2-bin \
           ghostscript \
-          graphviz \
           pkg-config \
           libgraphviz-dev \
           gsfonts \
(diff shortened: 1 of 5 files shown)


.github/actions/install-dependencies/action.yml
```diff
@@ -15,11 +15,18 @@ runs:
...
           gsfonts \
           libfreetype6-dev \
           libfontconfig1-dev
       # Explanation of some of these dependencies:
       # https://github.com/ImageMagick/ImageMagick/issues/374#issuecomment-279252866
+
+    - uses: conda-incubator/setup-miniconda@35d1405e78aa3f784fe3ce9a2eb378d5eeb62169 # v2.1.1
+      with:
+        channels: conda-forge
+        python-version: ${{ matrix.python-version || '' }}
```
oliver-sanders, line 28: I think we should switch to use the open miniforge or mambaforge offerings rather than the propitiatory miniconda.

From [the docs](https://github.com/conda-incubator/setup-miniconda#example-10-miniforge) it looks like this will do it.

```suggestion
        python-version: ${{ matrix.python-version || '' }}
        miniforge-version: latest
```

oliver-sanders (commented):
LGTM, can we confirm this version of Graphviz doesn't suffer from the overflowing graph problem we encountered before?

MetRonnie:
> LGTM, can we confirm this version of Graphviz doesn't suffer from the overflowing graph problem we encountered before?

Haven't had this problem locally at all, with the more recent versions of Graphviz

MetRonnie:
Test failure due to spellcheck

.github/actions/install-dependencies/action.yml
```diff
@@ -15,11 +15,22 @@ runs:
...
+        miniforge-version: latest
+
+    # use login shell for conda activation
+    - shell: bash -leo pipefail {0}
```
oliver-sanders, line 33: There is an `activate-environment:` configuration which presumably handles this?

MetRonnie: `activate-environment` is set to `test` by default - the `test` env gets activated for every step. However, the default shell is `bash --noprofile --norc -eo pipefail {0}` which prevents this activation. So you have to use a login shell instead.

See https://github.com/conda-incubator/setup-miniconda#important

PR #550, opened 2022-10-06
add spellings "len", "pyproject" and "toml" to dictionaries.

wxtim:
**Requirements check-list**

### DIFF
--- a/src/dictionaries/words
+++ b/src/dictionaries/words
@@ -68,6 +68,7 @@ iterable
 jitter
 jobscript
 latin
+len
 lifecycle
 linearization
 linter
--- a/src/dictionaries/proper_nouns
+++ b/src/dictionaries/proper_nouns
@@ -30,12 +30,14 @@ Posix
 Protobuf
 Pympler
 Pyro
+pyproject
 rsync
 sge
 slurm
 Slurm
 sqlite
 svn
+toml
 Trac
 Tui
 Vuex


src/dictionaries/words
```diff
@@ -68,6 +68,7 @@ iterable
 jitter
 jobscript
 latin
+len
```
wxtim, line 71: In the normal course of events I'd prefer this to be a case-by-case exception directive, but I don't want tot try and insert such a thing into lint which also prints the CLI reference.

oliver-sanders: Is `len` coming from a code example which hasn't been put in double backquotes? e.g:

```diff
- here's an example len(foo), this will
+ here's an example ``len(foo)``, this will
```

datamel:
Just query about the base branch here, is this a master change or 8.0.x? Not fully up to speed with what prs like this should go into.

MetRonnie:
The spelling problem is on 8.0.x so makes sense to put this against that IMO

src/dictionaries/words
wxtim: I think that you are right. Cylc PR to follow then. 🙄

wxtim:
```suggestion
```

wxtim:
This branch will now fail until https://github.com/cylc/cylc-flow/pull/5180 is merged.

MetRonnie:
> > The spelling problem is on 8.0.x (as well as master) so makes sense to put this against 8.0.x IMO
> 
> Why, the words have only been added to Cylc Master - Is the Cylc-Doc test using the wrong branch of Cylc?

Another complexity of having these parallel development branches... I see about fixing this

MetRonnie (commented):
`len` still there

MetRonnie (approved):

PR #553, opened 2022-10-10
Fix spellcheck on GH Actions

MetRonnie:
With the change from `ubuntu-18.04` to `ubuntu-latest`, the `aspell` language dictionaries were no longer installed, so the spellchecker started using `hunspell` instead, which seemed to be falling back on `en_US` despite us setting `en_NZ`.

**Requirements check-list**

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -109,10 +109,11 @@ jobs:
             html \
             slides \
             linkcheck \
-            SPHINXOPTS="-Wn -A sidebar_version_name=${version}" \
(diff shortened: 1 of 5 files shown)


.github/workflows/nightly.yml
```diff
@@ -109,10 +109,11 @@ jobs:
             html \
             slides \
             linkcheck \
-            SPHINXOPTS="-Wn -A sidebar_version_name=${version}" \
+            SPHINXOPTS="-Wn --keep-going -A sidebar_version_name=${version}" \
```
wxtim, line 112: Nice. If a little weird - one might assume that --keep-going is the opposite of -W.

oliver-sanders (approved):

PR #562, opened 2022-10-28
Get list of template variables for xtriggers from correct place

### LINKED ISSUE #546: Slight error in the xtrigger documentation around uuid?
I was looking at https://cylc.github.io/cylc-doc/latest/html/user-guide/writing-workflows/external-triggers.html#custom-trigger-functions and noticed mention of `uuid` and `suite_uuid`. The `suite_uuid` mentions it is deprecated and to use `uuid_str`. I'm not sure which is correct, but I'm guessing `uuid` should be `uuid_str` or `uuid_str` should be `uuid`?

image

wxtim:
Closes #546 

From cylc.flow.xtrigger_mgr.TemplateVariables, not task_events_mgr.EventData

# How to reproduce bug

```
#!jinja2
# flow.cylc
[scheduler]
    allow implicit tasks = True
[scheduling]
    initial cycle point = 2781-03-21T02:17
    [[graph]]
        R1 = @my_trigger => foo
    [[xtriggers]]
        my_trigger = my_trigger(%({{VAR}})s)
```
```
# lib/python/my_trigger.py
def my_trigger(*args):
    print(*args)
    return (True, {})
```
Running script:
```
cylc install -n foo && cylc play foo -s 'VAR="uuid"'
```
Should return an exception for variables documented which are not actually available as xtrigger template variables such as `suite_uuid` and `uuid`.

**Requirements check-list**

Rebuilt docs available on my internal web space with alleged version 8.99.0.dev

### DIFF
--- a/src/user-guide/writing-workflows/external-triggers.rst
+++ b/src/user-guide/writing-workflows/external-triggers.rst
@@ -254,7 +254,11 @@ properties:
    because each call is executed in an independent process in the process
    pool. If necessary the filesystem can be used for this purpose.
 
-.. autoenumvalues:: cylc.flow.task_events_mgr.EventData
+.. spelling:word-list::
+
+   vv
+
+.. autoenumvalues:: cylc.flow.xtrigger_mgr.TemplateVariables
 
 Function return values should be as follows:
 


MetRonnie (approved):
I'm not sure I understand your "how to reproduce". But I can clearly see the existing docs has the wrong source for the `autoenumvalues` directive, and this fixes it

wxtim:
> I'm not sure I understand your "how to reproduce". But I can clearly see the existing docs has the wrong source for the `autoenumvalues` directive, and this fixes it

That workflow allows you to try passing information to the x-trigger script which prints it using `cylc play -s 'VAR="<changeme>"'`.

You can double check which variables are correct by changing "changeme" for each of the documented variables.

oliver-sanders (approved):

PR #590, opened 2023-02-03
Document use of mamba to install Cylc.

hjoliver:
**Requirements check-list**

### DIFF
--- a/src/installation.rst
+++ b/src/installation.rst
@@ -15,22 +15,26 @@ Cylc runs on Unix-like systems including Linux and Mac OS.
 Via Conda (recommended)
 ^^^^^^^^^^^^^^^^^^^^^^^
 
-.. warning::
+.. note::
 
-   The ``conda`` environment solver can be very slow and memory-intensive for
-   complex packages. We recommend installing the much faster ``mamba`` solver
-   first, and then installing Cylc with that.
+   We recommend using the fast Mamba environment solver to install Cylc.
+   Mamba can be used as `a drop-in replacement for the conda command
+   <https://mamba.readthedocs.io/en/latest/index.html>`_,
+   or as `a conda command plugin
+   <https://conda.github.io/conda-libmamba-solver/getting-started/>`_.
+   The classic conda environment solver may be too slow for a complex package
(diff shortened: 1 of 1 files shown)


src/installation.rst
```diff
@@ -15,15 +15,22 @@ Cylc runs on Unix-like systems including Linux and Mac OS.
...
+.. warning::
+
+   The ``conda`` environment solver can be very slow and memory-intensive for
+   complex packages. We recommend installing the much faster ``mamba`` solver
+   first, and then installing Cylc with that.
+
 .. code-block:: sub
 
-   $ conda install -c conda-forge cylc-flow
+   $ conda install -c conda-forge mamba
```
oliver-sanders, line 26: This suggests installing `mamba` into the cylc env (or cylc into the base env).

oliver-sanders (commented):
The `mamba` solver is now available in `conda` too.

* https://www.anaconda.com/blog/a-faster-conda-for-a-growing-community
* https://conda.github.io/conda-libmamba-solver/getting-started/#getting-started

I think it still requires explicit installation/configuration but that may depend on where you got conda from.

Maybe replace this warning with a `.. note::` just saying "We recommend using `mamba`".

Mamba is the free as in speaking-chips replacement for Anaconda's free as in free until-you-cross-the-billing-threshold implementation which will hopefully supplant conda in the near future.

src/installation.rst
hjoliver: I see I have an extra erroneous line in there (fixed) ... but it is OK to install mamba into the cylc env.

hjoliver:
> The mamba solver is now available in conda too.

Nice, I wasn't aware of that.

hjoliver:
@oliver-sanders - I've tweaked the text along the lines you suggested. See what you think.

src/installation.rst
oliver-sanders: The docs says conda and mamba will only work when installed in the "base" env. You can install them into other envs but I guess they won't work for environment management, dunno.

oliver-sanders (approved):
Cheers.

PR #592, opened 2023-02-23
major changes: add parameter offsets page

oliver-sanders:
A change that we forgot about which was recently noticed: https://github.com/cylc/cylc-flow/pull/3452

**Requirements check-list**

### DIFF
--- a/src/7-to-8/major-changes/parameters.rst
+++ b/src/7-to-8/major-changes/parameters.rst
@@ -42,7 +42,7 @@ partial evaluation of that line:
    a_x1 => b_x2 => c_x2  # x=2
    a_x2 => b_x3 => c_x3  # x=3
 
-Whereas Cylc 8 will remove the entire line:
+Whereas Cylc 8 will remove everything after the first out-of-range parameter - `<x-1>` (so the entire line for this example):
 
 .. code-block:: cylc-graph
 
--- a/src/7-to-8/major-changes/parameters.rst
+++ b/src/7-to-8/major-changes/parameters.rst
@@ -1,0 +1,100 @@
+Parameters
+==========
+
+.. admonition:: does this change affect me?
+   :class: tip
+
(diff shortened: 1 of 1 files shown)


src/7-to-8/major-changes/parameters.rst
```diff
@@ -0,0 +1,100 @@
...
+           b_x1 => c_x1  # x=1
+   a_x1 => b_x2 => c_x2  # x=2
+   a_x2 => b_x3 => c_x3  # x=3
+
+Whereas Cylc 8 will remove the entire line:
+
+.. code-block:: cylc-graph
+
+   a_x1 => b_x2 => c_x2  # x=2
+   a_x2 => b_x2 => c_x2  # x=3
```
oliver-sanders, line 50: I'm struggling to remember this one it was so long ago.

Is this correct, or does Cylc remove the part of the line after the first out-of-bounds parameter?

src/7-to-8/major-changes/parameters.rst
```diff
@@ -0,0 +1,100 @@
...
+Cylc 7 removed the part of the expression which was out of range resulting in a
+partial evaluation of that line:
+
+.. code-block:: cylc-graph
+
+           b_x1 => c_x1  # x=1
+   a_x1 => b_x2 => c_x2  # x=2
+   a_x2 => b_x3 => c_x3  # x=3
+
+Whereas Cylc 8 will remove the entire line:
```
oliver-sanders, line 45: I think this is correct:

```suggestion
Whereas Cylc 8 will remove everything after the first out-of-range parameter - `<x-1>` (so the entire line for this example):
```

hjoliver: Yep.

src/7-to-8/major-changes/parameters.rst
hjoliver: (you got it right in the other comment).

hjoliver (approved):

PR #624, opened 2023-07-11
syntax: document cylc.vim plugin

oliver-sanders:
Document the cylc.vim plugin.

Merge once the plugin is ready to advertise.

The old syntax file to be removed with https://github.com/cylc/cylc-flow/issues/5626

**Requirements check-list**


hjoliver (approved):
I think this can be undrafted and merged now @oliver-sanders

PR #632, opened 2023-07-28
event handlers: improve docs

oliver-sanders:
**Requirements check-list**

Goes with https://github.com/cylc/cylc-flow/pull/5655

### DIFF
--- a/src/glossary.rst
+++ b/src/glossary.rst
@@ -1654,8 +1654,6 @@ Glossary
       WARNING messages from a task :term:`job`.
 
 
-   .. TODO cylc-flow cfgspec/workflow.py references "event handlers" plural
-
    handler
    event handler
    event handlers
@@ -1671,9 +1669,8 @@ Glossary
 
       .. seealso::
 
-         - :cylc:conf:`task events <[runtime][<namespace>][events]>`
-         - :cylc:conf:`workflow events <[scheduler][events]>`
-         - :ref:`Cylc User Guide <EventHandling>`
+         - :ref:`user_guide.scheduler.workflow_event_handling`
+         - :ref:`user_guide.runtime.task_event_handling`
 
 
    runahead limit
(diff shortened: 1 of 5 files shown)


src/user-guide/writing-workflows/runtime.rst
```diff
@@ -634,18 +637,40 @@ can be:
...
-are triggered.
+.. cylc-scope:: flow.cylc[runtime][<namespace>][events]
+
+Alternatively you can configure a list of generic event :cylc:conf:`handlers` to be run
+for configured :cylc:conf:`handler events`.
+
+:cylc:conf:`handler events`
+   A list of events which may include any of the above
+   events (e.g. ``submission failed`` or ``warning``) or
+   any of a task's :term:`custom outputs <custom output>`.
```
oliver-sanders, line 654: This is piece of information which was missing before.

Turns out custom outputs can be used in `handler events`, this is covered by an existing functional test.

src/glossary.rst
```diff
@@ -1654,8 +1654,6 @@ Glossary
       WARNING messages from a task :term:`job`.
 
 
-   .. TODO cylc-flow cfgspec/workflow.py references "event handlers" plural
```
oliver-sanders, line 1657: The `event handlers` config is plural so this seems legit to me.

wxtim (approved):
Yes please - I find these settings slightly hard to think about and this documentation update helps.


Probably out of scope...
 I think that there is ambiguity because "handlers" is plural because it might be running multiple scripts, whilst "event handlers" is plural because it might run on multiple events. I wonder if `common handlers` and `common events` might be better terms. Or at least `events handlers`?

oliver-sanders:
I think we might consider a rethink of how we configure event handlers in light of https://github.com/cylc/cylc-flow/issues/5652.

It would make more sense to configure events and handlers as groups e.g. for "failed and submit-failed" events, run handlers x, y, & z, but for "succeeded and info" events, run handler a.

PR #636, opened 2023-08-22
Exclude broken sphinx version

MetRonnie:
**Requirements check-list**

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -51,7 +51,7 @@ install_requires =
     eralchemy==1.2.*
     hieroglyph>=2.1.0
     setuptools>=50
-    sphinx>=4.4
+    sphinx>=4.4,!=7.2.*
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -51,7 +51,7 @@ install_requires =
     eralchemy==1.2.*
     hieroglyph>=2.1.0
     setuptools>=50
-    sphinx>=4.4
+    sphinx>=4.4,!=7.2.*
```
oliver-sanders, line 54: Are we expecting this issue to be fixed in 7.3?

MetRonnie: Don't know when it will be fixed, but hopefully? I've opened an issue: https://github.com/sphinx-doc/sphinx/issues/11631

oliver-sanders (approved):

PR #638, opened 2023-08-29
troubleshooting

oliver-sanders:
The beginnings of a Cylc 8 troubleshooting section, mostly aimed at users as opposed to sys-admins.

This is on an upstream branch, please contribute!

Specific troubleshooting for platform setup issues (sys-admin targetted content) is probably better placed in `src/reference/config/writing-platform-configs.rst`.

**Requirements check-list**

### DIFF
--- a/src/troubleshooting.rst
+++ b/src/troubleshooting.rst
@@ -1,0 +1,204 @@
+Troubleshooting
+===============
+
+If things have gone wrong and you're not sure why, there are few files which
+should contain the required information to work out what's going on.
+
+``log/scheduler/log``
+   There's a log file for each workflow in
+   ``~/cylc-run/<workflow>/log/scheduler/log``.
+
+   You can view this in the GUI or on the command line using
+   ``cylc cat-log <workflow>``.
+``job.err``
+   This contains the stderr captured when the job ran. It's useful for
+   debugging job failures.
+
+   You can view this in the GUI or on the command line using
+   ``cylc cat-log <workflow>//<cycle>/<task>/<job> -f e``.
+``job-activity.log``
+   This file records the interaction Cylc has had with a job, namely submission
+   and polling. This can be useful in determining the cause of job submission
+   failures.
+
+   You can view this in the GUI or on the command line using
+   ``cylc cat-log <workflow>//<cycle>/<task>/<job> -f a``.
+
+
+Problems
+--------
+
+
+Job Status Isn't Updating
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Cylc keeps track of a job's progress in one of two ways (according to how
+the platform the job was submitted to is configured):
+
+* Jobs send messages to the scheduler (push).
+* The scheduler polls jobs (pull).
+
+In either case, the job will also write its updates to the ``job.status`` file.
+
+This is what the ``job.status`` file should look like for a successful job,
+note the ``SUCCEEDED`` line:
+
+.. code-block::
+
+   CYLC_JOB_RUNNER_NAME=background
+   CYLC_JOB_ID=12345
+   CYLC_JOB_RUNNER_SUBMIT_TIME=2000-01-01T00:00:00
+   CYLC_JOB_PID=108179
+   CYLC_JOB_INIT_TIME=2000-01-01T00:10:00
+   CYLC_JOB_EXIT=SUCCEEDED
+   CYLC_JOB_EXIT_TIME=2000-01-01T01:30:00
+
+If the ``job.status`` file is showing something different to what the GUI or
+Tui is showing, then...
+
+.. rubric:: If your platform uses push communication:
+
+If messages aren't getting back to the scheduler, there should be some
+evidence of this in the ``job.err`` file, likely either an error or a
+traceback.
+
+Likely causes:
+
+* There is a network issue.
+* TCP ports are not open (zmq communications).
+* Non-interactive SSH has not been correctly configured (ssh communications).
+
+.. rubric:: If your platform uses pull communication:
+
+Firstly, check the polling interval, it's possible that the scheduler has been
+configured to poll infrequently and you need to wait for the next poll, or use
+the ``cylc poll`` command (also available in the GUI).
+
+Use the ``cylc config`` command to inspect the platform's configuration to
+determine the configured polling schedule.
+
+Then check the ``job-activity.log`` file, there may have been a problem polling
+the remote platform, e.g. a network or configuration error.
+
+Likely causes:
+
+* The platform is down (e.g. all login nodes are offline).
+* There is a network issue.
+* Non-interactive SSH has not been correctly configured.
+
+
+My Job Submit-Failed
+^^^^^^^^^^^^^^^^^^^^
+
+A submit-failed job means one of three things:
+
+1. There is a Bash syntax error in the task configuration.
+
+   E.G. the following ``script`` has a syntax error, it is missing a
+   ``"`` character:
+   
+   .. code-block:: cylc
+   
+      [runtime]
+          [[foo]]
+              script = echo "Hello $WORLD
+   
+   This will result in a submission-failure which should appear in the
+   ``job-activity.log`` file (and also the scheduler log) something like this:
+   
+   .. code-block::
+   
+      /path/to/job.tmp: line 46: unexpected EOF while looking for matching `"'
+      /path/to/job.tmp: line 50: syntax error: unexpected end of file
+
+2. There was an error submitting the job to the specified platform (including
+   network issues).
+
+   See the ``job-activity.log`` and the scheduler log. The error should be in
+   one or both of those files.
+
+3. The platform is not correctly configured.
+
+
+My Job Failed
+^^^^^^^^^^^^^
+
+This means something went wrong executing the job.
+
+To find out more, see the ``job.err`` file.
+
+If you're struggling to track down the error, you might want to restart the
+workflow in debug mode and run the task again:
+
+.. code-block:: console
+
+   # shut the workflow down (leave any active jobs running)
+   $ cylc stop --now --now <workflow>
+   # restart the workflow in debug mode
+   $ cylc play <workflow> --debug
+   # re-run all failed task(s)
+   $ cylc trigger '<workflow>//*:failed'
+
+When a workflow is running in debug mode, all jobs will create a ``job.xtrace``
+file which can help you to locate the error within the job script.
+
+
+My workflow shutdown unexpectedly
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+When a Cylc scheduler shuts down, it should leave behind a log message explaining why.
+
+E.G. this message means that a workflow shut down because it was told to:
+
+.. code-block::
+
+   Workflow shutting down - REQUEST(CLEAN)
+
+If a workflow shut down due to a critical problem, you should find some
+traceback in this log. If this traceback doesn't look like it comes from your
+system, please report it to the Cylc developers for investigation (on
+GitHub or Discourse).
+
+In some extreme cases, Cylc might not be able to write a log message e.g:
+
+* There's not enough disk space for Cylc to write a log message.
+* If the scheduler will killed in a nasty way e.g. ``kill -9``.
+* If the scheduler host goes down (e.g. power off).
+
+
+Error Messages
+--------------
+
+
+FileNotFoundError: No such file or directory
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This is the error message Python gives when you try to call an exectuable which
+does not exist in the ``$PATH``. It means there's something wrong with the Cylc
+installation.
+
+E.G. the following error:
+
+.. code-block::
+
+   FileNotFoundError: [Errno 2] No such file or directory: 'ssh'
+
+Means that ``ssh`` is not installed.
+
+See :ref:`non-python-requirements` for details on system requirements.
+
+
+platform: <name> - initialisation did not complete
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This means that Cylc was unable to install the required workflow files onto
+a remote platform.
+
+This either means that:
+
+1. The platform is down (e.g. all login nodes are offline).
+2. There is a network problem (e.g. you cannot connect to the login nodes).
+3. The platform is not correctly configured.
+
+Check the scheduler log, you might find some stderr associated with this
+message.
--- a/src/troubleshooting.rst
+++ b/src/troubleshooting.rst
@@ -225,36 +225,32 @@ To find out more, see the ``job.err`` file.
 
    .. cylc-scope::
 
-If you're struggling to track down the error, you might want to restart the
-workflow in debug mode and run the task again:
+If you're struggling to track down the error, you might want to put the
+workflow into debug mode::
 
-.. TODO Update this advice after https://github.com/cylc/cylc-flow/issues/5829
-
-.. code-block:: console
-
-   # shut the workflow down (leave any active jobs running)
-   $ cylc stop --now --now <workflow>
-   # restart the workflow in debug mode
-   $ cylc play <workflow> --debug
-   # re-run all failed task(s)
-   $ cylc trigger '<workflow>//*:failed'
+   cylc verbosity DEBUG <workflow-id>
 
 When a workflow is running in debug mode, all jobs will create a ``job.xtrace``
-file which can help you to locate the error within the job script.
+file when run in addition to ``job.err``. This can help you to locate the error
+within the job script.
+
+You can also start workflows in debug mode::
+
+   cylc play --debug <workflow-id>
 
 
-My workflow shutdown unexpectedly
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+My workflow shut down unexpectedly
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 When a Cylc scheduler shuts down, it should leave behind a log message explaining why.
 
-E.G. this message means that a workflow shutdown because it was told to:
+E.G. this message means that a workflow shut down because it was told to:
 
 .. code-block::
 
    Workflow shutting down - REQUEST(CLEAN)
 
-If a workflow shutdown due to a critical problem, you should find some
+If a workflow shut down due to a critical problem, you should find some
 traceback in this log. If this traceback doesn't look like it comes from your
 system, please report it to the Cylc developers for investigation (on
 GitHub or Discourse).
@@ -276,6 +272,10 @@ Why isn't my task running?
 To find out why a task is not being run, use the ``cylc show`` command.
 This will list the task's prerequisites and xtriggers.
 
+Note, at present ``cylc show`` can only display
+:term:`active tasks <active task>`. Waiting tasks beyond the
+:term:`n=0 window <n-window>` have no satisfied prerequisites.
+
 Note, tasks which are held |task-held| will not be run, use ``cylc release``
 to release a held task.
 
@@ -296,7 +296,7 @@ If something has gone wrong during installation, an error should have been
 logged a file in this directory:
 ``$HOME/cylc-run/<workflow-id>/log/remote-install/``.
 
-``If you need to access files from a remote platform (e.g. 2-stage ``fcm_make``),
+If you need to access files from a remote platform (e.g. 2-stage ``fcm_make``),
 ensure that a task has submitted to it before you do so. If needed you can use
 a blank "dummy" task to ensure that remote installation is completed *before*
 you run any tasks which require this e.g:
@@ -312,12 +312,12 @@ Conda / Mamba environment activation fails
 Some Conda packages rely on activation scripts which are run when you call the
 activate command.
 
-Sadly, some of these scripts don't defend against command failure or unset
-environment variables causing them to fail when configured in Cylc ``*script``
-(see also :ref:`troubleshooting.my_job_failed` for details).
+Unfortunately, some of these scripts don't defend against command failure or
+unset environment variables causing them to fail when configured in Cylc
+``*script`` (see also :ref:`troubleshooting.my_job_failed` for details).
 
 To avoid this, run ``set +eu`` before activating your environment. This turns
-off some Bash safety features allowing environment activation to complete.
+off some Bash safety features, allowing environment activation to complete.
 Remember to run ``set -eu`` afterwards to turn these features back on.
 
 .. code-block:: cylc
@@ -360,7 +360,7 @@ E.G. the following error:
 
    FileNotFoundError: [Errno 2] No such file or directory: 'ssh'
 
-Means that ``ssh`` is not installed.
+Means that ``ssh`` is not installed or not in your ``$PATH``.
 
 See :ref:`non-python-requirements` for details on system requirements.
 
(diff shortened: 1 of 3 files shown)


src/troubleshooting.rst
```diff
@@ -0,0 +1,204 @@
...
+   # shut the workflow down (leave any active jobs running)
+   $ cylc stop --now --now <workflow>
+   # restart the workflow in debug mode
+   $ cylc play <workflow> --debug
+   # re-run all failed task(s)
+   $ cylc trigger '<workflow>//*:failed'
+
+When a workflow is running in debug mode, all jobs will create a ``job.xtrace``
+file which can help you to locate the error within the job script.
+
```
wxtim, line 144: Worth shoehorning a note about `set -eu` in here? Something like:

```suggestion
.. note::

   "My job didn't fail, but should have": 
   
   Bash scripts do not by default fail on error. Use `set -e` to alter this default. 
   
   See :ref:`workflow_design_guide.basic_principles.basic_functionality`

```

src/troubleshooting.rst
```diff
@@ -0,0 +1,204 @@
...
+          [[foo]]
+              script = echo "Hello $WORLD
+   
+   This will result in a submission-failure which should appear in the
+   ``job-activity.log`` file (and also the scheduler log) something like this:
+   
+   .. code-block::
+   
+      /path/to/job.tmp: line 46: unexpected EOF while looking for matching `"'
+      /path/to/job.tmp: line 50: syntax error: unexpected end of file
```
hjoliver, line 112:
```suggestion
      [jobs-submit cmd] (prepare job file)
      [jobs-submit ret_code] 1
      [jobs-submit err]
      /path/to/job.tmp: line 46: unexpected EOF while looking for matching `"'
      /path/to/job.tmp: line 50: syntax error: unexpected end of file
      
Cylc checks job syntax just before submission (because active broadcasts may alter task scripting on the fly) and aborts the process if errors are found.
```
I think it's worth adding more context here - it shows that job preparation failed, not execution - which might be expected for a bash syntax error.

src/troubleshooting.rst
```diff
@@ -0,0 +1,204 @@
...
+
+2. There was an error submitting the job to the specified platform (including
+   network issues).
+
+   See the ``job-activity.log`` and the scheduler log. The error should be in
```
hjoliver, line 117:
```suggestion
   See the ``job-activity.log`` and the scheduler log. Diagnostic information should be in
```

src/troubleshooting.rst
hjoliver: (Should mention that Cylc job scripts do use `set -eu`, so this advice applies to external scripts called by the job)

src/troubleshooting.rst
```diff
@@ -0,0 +1,204 @@
...
+   $ cylc play <workflow> --debug
+   # re-run all failed task(s)
+   $ cylc trigger '<workflow>//*:failed'
+
+When a workflow is running in debug mode, all jobs will create a ``job.xtrace``
+file which can help you to locate the error within the job script.
+
+
+My workflow shutdown unexpectedly
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
hjoliver, line 147:
```suggestion
My workflow shut down unexpectedly
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

src/troubleshooting.rst
```diff
@@ -0,0 +1,204 @@
...
+
+If a workflow shut down due to a critical problem, you should find some
+traceback in this log. If this traceback doesn't look like it comes from your
+system, please report it to the Cylc developers for investigation (on
+GitHub or Discourse).
+
+In some extreme cases, Cylc might not be able to write a log message e.g:
+
+* There's not enough disk space for Cylc to write a log message.
+* If the scheduler will killed in a nasty way e.g. ``kill -9``.
```
hjoliver, line 165:
```suggestion
* If the scheduler was killed in a nasty way e.g. ``kill -9``.
```

src/troubleshooting.rst
```diff
@@ -0,0 +1,204 @@
...
+* If the scheduler host goes down (e.g. power off).
```
hjoliver, line 166:
```suggestion
* If the scheduler host goes down (e.g. power off). 
  (Note if this happens you can simply play the workflow again once the host is back up - it will start up from where it left off when the host went down)
```

src/troubleshooting.rst
```diff
@@ -0,0 +1,204 @@
...
+
+
+Error Messages
+--------------
+
+
+FileNotFoundError: No such file or directory
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This is the error message Python gives when you try to call an exectuable which
```
hjoliver, line 176:
```suggestion
This is the error message Python gives when you try to call an executable which
```

src/troubleshooting.rst
```diff
@@ -0,0 +1,204 @@
...
+does not exist in the ``$PATH``. It means there's something wrong with the Cylc
+installation.
```
hjoliver, line 178:
```suggestion
does not exist in the ``$PATH``. It means there's something wrong with the Cylc
installation or your default environment for non-interactive shells.
```

src/troubleshooting.rst
```diff
@@ -0,0 +1,204 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This means that Cylc was unable to install the required workflow files onto
+a remote platform.
+
+This either means that:
+
+1. The platform is down (e.g. all login nodes are offline).
+2. There is a network problem (e.g. you cannot connect to the login nodes).
+3. The platform is not correctly configured.
```
hjoliver, line 201:
```suggestion
3. The platform is not correctly configured.  For example, you have not specified a valid install target for Cylc to install workflow files on the platform.
```

hjoliver (commented):
Good.

> This is on an upstream branch, please contribute!

Erm, what do you mean by this?  Are you suggesting we don't just merge individual PRs to master as usual, to incrementally upgrade the troubleshooting section?

src/troubleshooting.rst
```diff
@@ -0,0 +1,204 @@
...
+1. There is a Bash syntax error in the task configuration.
+
+   E.G. the following ``script`` has a syntax error, it is missing a
+   ``"`` character:
+   
+   .. code-block:: cylc
+   
+      [runtime]
+          [[foo]]
+              script = echo "Hello $WORLD
```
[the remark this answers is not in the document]
wxtim: Surely the `"Hello $WORLD` out to have a closing parenthesis?

oliver-sanders:
> Good.
>
>> This is on an upstream branch, please contribute!
>
> Erm, what do you mean by this? Are you suggesting we don't just merge individual PRs to master as usual, to incrementally upgrade the troubleshooting section?

The branch this PR is on is a cylc/cylc-doc branch NOT an oliver-sanders/cylc-doc branch, that means you can push new commits to this PR. I.E. I have intended this PR to be collaborative.

src/troubleshooting.rst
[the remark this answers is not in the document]
hjoliver: He's giving an example of a bash syntax error that causes job submit failure. 

My point is, to get that far it needs to pass validation first.

hjoliver:
> The branch this PR is on is a cylc/cylc-doc branch NOT an oliver-sanders/cylc-doc branch, that means you can push new commits to this PR. I.E. I have intended this PR to be collaborative.

Roger that (well, I think we tend to allow pushes to each other's branch too, but I guess a central branch is more clearly for that purpose).

oliver-sanders:
Should add "Disk IO Error" (ran out of disk space)

src/troubleshooting.rst
oliver-sanders: Nope, we've added `shutdown` into src/dictionaries/words.

src/troubleshooting.rst
oliver-sanders: This should be true for all cases, but probably isn't, so leaving it out for now.

src/troubleshooting.rst
oliver-sanders: I don't think that should be relevant here (it may also be inaccurate, e.g. my default environment ... might be ksh not Bash, but Cylc won't use that).

I think what's written covers "you need to install <x>".

oliver-sanders:
Addressed feedback, fleshed this out a bit.

I think there's enough here to make it worthwhile merging this now so changed the milestone.

oliver-sanders:
(rebased)

wxtim (approved):
Looks good. I was going to add a couple of things, but it might be better to get it in then I'll propose those additions.

oliver-sanders:
For the record my other thoughts were:

>    Did your script fail because you needed to chmod?

Good idea.

>    log/config/flow-processed.cylc shows you the effect of Jinja2 templating.

Good idea, we could do with a general Jinja2 debugging section including `cylc view -p` and `flow-processed.cylc`.

>   log/config/01-start-01.cylc shows you the effect of Jinja2 and task parameter expansions.

I'm not sure we expand task params, or maybe just in runtime? Can't remember. Good to raise awareness of the config log directory, explain the file numbers and names and why they are useful.

wxtim:
> I'm not sure we expand task params, or maybe just in runtime?

I wasn't sure, so I checked - and can justify this statement.

src/troubleshooting.rst
hjoliver: I'm not saying "shutdown" is not an accepted word. It's just not correct in this context.

- "shutdown" is a noun - my scheduler suffered an unexpected shutdown
- "shut down" is verb - my scheduler is about to shut down

E.g., microsoft style guide:
> https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/s/shut-down-shutdown

src/troubleshooting.rst
```diff
@@ -0,0 +1,482 @@
...
+does not exist in the ``$PATH``. It means there's something wrong with the Cylc
+installation.
+
+E.G. the following error:
+
+.. code-block::
+
+   FileNotFoundError: [Errno 2] No such file or directory: 'ssh'
+
+Means that ``ssh`` is not installed.
```
hjoliver, line 363:
```suggestion
Means that ``ssh`` is not installed or not in your ``$PATH``.
```

Or perhaps:
```suggestion
Means that ``ssh`` is not installed or not accessible in your environment.
```

src/troubleshooting.rst
```diff
@@ -0,0 +1,482 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This means that Cylc was unable to install the required workflow files onto
+a remote platform.
+
+This either means that:
+
+1. The platform is down (e.g. all login nodes are offline).
+2. There is a network problem (e.g. you cannot connect to the login nodes).
+3. The platform is not correctly configured.
```
hjoliver, line 380: Use of "either" requires "or" as well.

```suggestion
This either means that:

1. The platform is down (e.g. all login nodes are offline)
2. Or, there is a network problem (e.g. you cannot connect to the login nodes)
3. Or, the platform is not correctly configured.
```

src/troubleshooting.rst
```diff
@@ -0,0 +1,482 @@
...
+You might want to check that you've configured the
+:cylc:conf:`global.cylc[platforms][<platform name>]install target` correctly,
+note that this defaults to the platform name if not explicitly set.
+
+
+``OperationalError: disk I/O error``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This means that something was unable to write to a file when it would expect to
+have been able to.
```
hjoliver, line 399: I think you're trying to avoid mention the DB to users here?  However, that's an sqlite-specific error, and with disk quota exceeded they might well get other less-specific IO errors first (or as well). I think it would be fine to mention the workflow DB here - no need to explain exactly what it's for.

src/troubleshooting.rst
```diff
@@ -0,0 +1,482 @@
...
+
+This error usually occurs if when you have exceeded you filesystem quota.
```
hjoliver, line 401:
```suggestion
This error usually occurs if when you have exceeded your filesystem quota.
```

src/troubleshooting.rst
```diff
@@ -0,0 +1,482 @@
...
+
+If a Cylc workflow cannot write to the filesystem, it will shutdown. Once
```
hjoliver, line 403:
```suggestion
If a Cylc scheduler cannot write to the filesystem, it will shutdown. Once
```
This has wider implications, so could leave for later. We've all long been using "scheduler" and "workflow" (meaning "running workflow") interchangeably, but I'd prefer not to conflate the two fundamentally different concepts.

src/troubleshooting.rst
```diff
@@ -0,0 +1,482 @@
...
+you've cleared out enough space for the workflow to continue you should be able
+to safely restart it as you would normally using ``cylc play``, the workflow
```
hjoliver, line 405:
```suggestion
to safely restart it as you would normally using ``cylc play``. The workflow
```

src/troubleshooting.rst
```diff
@@ -0,0 +1,482 @@
...
+will continue where it left off.
```
hjoliver, line 406:
```suggestion
will continue from where it left off.
```

src/troubleshooting.rst
```diff
@@ -0,0 +1,482 @@
...
+This usually means that a host could not be found on the network. The likely
+cause is DNS configuration.
+
+Cylc is a distributed system so needs to be able to identify the hosts it has
+been configured to use (e.g. the servers where you run Cylc workflows or any
+login nodes you submit jobs to).
+Cylc expects each host to have a unique and stable fully qualified domain name
+(FQDN) and to be identifiable from other hosts on the network using this name.
+
+I.E. If a host identifies itself with an FQDN, then we should be able to look it
```
hjoliver, line 421:
```suggestion
I.e., if a host identifies itself with an FQDN, then we should be able to look it
```

src/troubleshooting.rst
```diff
@@ -0,0 +1,482 @@
...
+
+If the FQDN of a host is reported differently from different hosts on the
+network, then Cylc commands will likely fail. To fix the issue, ensure that the
+DNS setup is consistent.
+
+Sometimes we do not have control over the platforms we use and it is not
+possible to compel system administrators to address these issues. If this is
+the case, you can fall back to IP address based host identification which may
+work (i.e. use IP address rather than host names, makes logs less human
+readable). As a last resort you can also hard-code the host name for each host.
```
hjoliver, line 433:
```suggestion
work (i.e. use IP addresses rather than host names, which makes logs less 
readable). As a last resort you can also hard-code the host name for each host

File: src/troubleshooting.rst [hunk 20]
```diff
@@ -0,0 +1,482 @@
...
+
+
+``Cannot determine whether workflow is running on <host>``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+When a Cylc workflow runs, it creates a :term:`contact file` which tells us on
+which host and port it can be contacted.
+
+If the workflow cannot be contacted, Cylc will attempt to check whether the
+process is still running to ensure it hasn't crashed.
```
hjoliver, line 456:
```suggestion
When Cylc runs a workflow, it creates a :term:`contact file` which tells us on
which host and port it can be contacted.

If the scheduler cannot be contacted, Cylc will attempt to check whether the
process is still running to ensure it hasn't crashed.
```
(Just another example of my complaint above - I think "contacting *the workflow*" is a bit too loose.

src/troubleshooting.rst
```diff
@@ -0,0 +1,482 @@
...
+
+If you are seeing this error message, it means that Cylc was unable to
+determine whether the workflow is running. Likely cause:
```
hjoliver, line 459: ("the workflow is running" is fine though)

hjoliver (changes requested):
Definitely need this in 8.3.0

src/troubleshooting.rst
```diff
@@ -0,0 +1,482 @@
...
+When a workflow is running in debug mode, all jobs will create a ``job.xtrace``
+file which can help you to locate the error within the job script.
+
+
+My workflow shutdown unexpectedly
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+When a Cylc scheduler shuts down, it should leave behind a log message explaining why.
+
+E.G. this message means that a workflow shutdown because it was told to:
```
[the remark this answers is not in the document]
hjoliver: (See above https://github.com/cylc/cylc-doc/pull/638#discussion_r1583888848 before responding to this)

hjoliver:
@wxtim - do you want to (or have you) posted another issue for your additional points above? https://github.com/cylc/cylc-doc/pull/638#pullrequestreview-1851388590

src/troubleshooting.rst
```diff
@@ -0,0 +1,482 @@
...
+   dummy => fetch_files
+
+
+Conda / Mamba environment activation fails
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Some Conda packages rely on activation scripts which are run when you call the
+activate command.
+
+Sadly, some of these scripts don't defend against command failure or unset
```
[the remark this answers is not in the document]
wxtim: Sadly: Over-not sure suitable for style
Unfortunately: Very long.

My suggestion: Don't make a value judgement, Skip the adjective.

src/troubleshooting.rst
wxtim: correct-futurama

On the other hand you could just skip the Latin _id est_ and say:

```suggestion
That is, if a host identifies itself with an FQDN, then we should be able to look it
```

src/troubleshooting.rst
```diff
@@ -0,0 +1,482 @@
...
+``socket.gaierror``
+^^^^^^^^^^^^^^^^^^^
+
+This usually means that a host could not be found on the network. The likely
+cause is DNS configuration.
+
+Cylc is a distributed system so needs to be able to identify the hosts it has
+been configured to use (e.g. the servers where you run Cylc workflows or any
+login nodes you submit jobs to).
+Cylc expects each host to have a unique and stable fully qualified domain name
```
wxtim, line 418: Consider

```suggestion
Cylc expects each host to have a unique and stable Fully Qualified Domain Name
```

src/troubleshooting.rst
hjoliver: Agreed, at the start of a sentence, needing capitalization, "I.e." doesn't read well.

wxtim:
> @wxtim - do you want to (or have you) posted another issue for your additional points above? [#638 (review)](https://github.com/cylc/cylc-doc/pull/638#pullrequestreview-1851388590)

It's #725

src/troubleshooting.rst
```diff
@@ -0,0 +1,482 @@
...
+If the issue is external to the workflow, once the issue is resolved it should
+be possible to restart it as you would normally using ``cylc play``. Cylc
+will pick up where it left off.
+
+
+Why isn't my task running?
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+To find out why a task is not being run, use the ``cylc show`` command.
+This will list the task's prerequisites and xtriggers.
```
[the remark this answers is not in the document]
oliver-sanders: Added note to remove this on the "task matching" issue - https://github.com/cylc/cylc-flow/issues/5827

src/troubleshooting.rst
```diff
@@ -0,0 +1,482 @@
...
+submits to it.
+
+See :ref:`RemoteInit` for the list of directories installed and how to
+configure them.
+
+If something has gone wrong during installation, an error should have been
+logged a file in this directory:
+``$HOME/cylc-run/<workflow-id>/log/remote-install/``.
+
+``If you need to access files from a remote platform (e.g. 2-stage ``fcm_make``),
```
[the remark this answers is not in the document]
oliver-sanders: The files aren't necessarily workflow files.

`fcm_make` is given as an example, this has hit a lot of workflows so it is reasonable to keep the reference.

src/troubleshooting.rst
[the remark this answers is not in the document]
oliver-sanders: "Value judgement" is beneficial here, it emphasises that Cylc isn't being awkward here.

I think this level of language fiddling is a bit too nit-picky, but changed to unfortunately anyway.

src/troubleshooting.rst
```diff
@@ -0,0 +1,482 @@
...
+If you are presented with traceback when you are *not* running in debug mode,
+then this is not an expected error, please report the traceback to us.
+
+
+``FileNotFoundError: No such file or directory``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This is the error message Python gives when you try to call an executable which
+does not exist in the ``$PATH``. It means there's something wrong with the Cylc
+installation.
```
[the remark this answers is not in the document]
oliver-sanders: SSH is a dependency of Cylc, so something is wrong with the Cylc installation.

I'm happy with this, if you are not please provide an alternative.

src/troubleshooting.rst
oliver-sanders: meh

oliver-sanders:
Comments addressed bar a couple.

src/troubleshooting.rst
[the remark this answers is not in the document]
hjoliver: I kinda disagree - the Cylc user guide shouldn't really reference undefined terms. But no biggie.

src/troubleshooting.rst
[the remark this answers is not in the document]
hjoliver: > I think this level of language fiddling is a bit too nit-picky,

Meh, that's what style guides for written English are for. Using "excessive informality" (as @wxtim put it somewhere else recently?) in technical docs stands out like a sore thumb and distracts from the content.

src/troubleshooting.rst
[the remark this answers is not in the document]
hjoliver: But it's not a dependency that can be satisfied by "installing Cylc" correctly.  Reductio ad absurdum, but could I install Cylc on MS Windows and claim that there is something wrong with the Cylc installation?

How about this then?
```suggestion
does not exist in the ``$PATH``. It means there's something wrong with the Cylc
installation, or something missing from the environment or system in which Cylc has been installed.
```

src/troubleshooting.rst
```diff
@@ -0,0 +1,481 @@
...
+
+
+``OperationalError: disk I/O error``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This means that Cylc was unable to write to the database.
+
+This error usually occurs if when you have exceeded your filesystem quota.
+
+If a Cylc scheduler cannot write to the filesystem, it will shutdown. Once
```
hjoliver, line 402:
```suggestion
If a Cylc scheduler cannot write to the filesystem, it will shut down. Once
```

(Last one of these)

src/troubleshooting.rst
wxtim: I went and looked it up. I'm wrong about this.

hjoliver (approved):

src/troubleshooting.rst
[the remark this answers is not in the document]
hjoliver: Update: `cylc show` now works via the datastore, on tasks in  the `n-window`.

PR #639, opened 2023-08-29
Update outdated UTC mode docs

MetRonnie:
`UTC mode` only affects log timestamp time zone. The important setting is `cycle point time zone`, which is `Z` be default.

**Requirements check-list**


hjoliver (approved):

MetRonnie:
1 reviewer should do. Just going to kick the tests...

PR #644, opened 2023-09-03
Update scheduling.rst

kinow:
rst doesn't support nesting formatting styles, so foo gets displayed with the double back ticks showing.

Happy to amend it to move foo inside the italic block if that's better.

**Requirements check-list**


oliver-sanders:
Hi @kinow :wave:.

Cheers.

oliver-sanders (approved):

PR #651, opened 2023-09-20
Explain code in flow.cylc files.

hjoliver:
Explain exactly what Jinja2 and shell code embedded in a `flow.cylc` means, and when/where it will be evaluated.

**Requirements check-list**

### DIFF
--- a/src/user-guide/writing-workflows/configuration.rst
+++ b/src/user-guide/writing-workflows/configuration.rst
@@ -62,10 +62,137 @@ documentation, configuration files, etc. When the workflow is :ref:`installed
    equivalent directories in the :ref:`workflow_share_directories`.
 
 
+.. _CodeInCylcConfigurations:
+
+Code in Workflow Configurations
+-------------------------------
+
+Cylc workflow configurations are not executable scripts or programs. Rather,
+they configure, at start-up, the Cylc scheduler program to run a particular
+workflow.
+
+The scheduler understands a static syntax that contains only ``[section
+headings]`` and configuration items of the form ``platform = hpc1``. These
+configuration items are not even "variables" that can be referenced elsewhere
+in the file. Some values may look complicated, but they are still just text
+strings with content that is meaningful to the scheduler.
+
+Nevertheless a `flow.cylc` file may appear to contain several kinds of embedded
+code:
+
+    - Python-like Jinja2 or EmPy code, such as ``{% set PLANET = "earth" %}``
+    - Bash shell variable assignments such as ``PLANET=${PLANET:-earth}``
+    - Bash shell scripting, such as ``script = "run-model.exe /path/to/data"``
+
+The documentation below explains exactly what these mean in the context of a
+workflow configuration, and when the code is evaluated.
+
+
+Jinja2 (or EmPy) Code
+^^^^^^^^^^^^^^^^^^^^^
+
+If a `flow.cylc` file contains embedded blocks of Jinja2 code in curly
+brackets (or the equivalent for EmPy) then it is actually a template for
+programmatically generating the static configuration format required by the
+scheduler.
+
+The template processor manipulates the surrounding text (which to it is
+entirely arbitrary) and the result must be valid configuration syntax that
+will pass validation by Cylc.
+
+This is a preprocessing step. Jinja2 code does not execute as the workflow runs.
+
+Use ``cylc view -j`` to see the result, with all Jinja2 code "processed out".
+
+Note that the template processor does not have access to config items from the
+surrounding text:
+
+.. code-block:: cylc
+
+   # flow.cylc
+   platform = hpc1  # Cylc config item definition
+   {{ "Platform is: " ~ platform }}  # ERROR, platform not defined as Jinja2!
+
+However, you can assign a value to a Jinja2 variable and print it to the file
+wherever it is needed:
+
+.. code-block:: cylc
+
+   # flow.cylc
+   {% set PLATFORM = "hpc1" %}
+   platform = {{ PLATFORM }}  # OK
+   {{ "Platform is: " ~ PLATFORM }} # OK
+ 
+Jinja2 code does not have access to the environment via normal shell syntax
+either, but you can read the local environment with native Jinja2 syntax:
+
+.. code-block:: cylc
+
+   # flow.cylc
+   {{ $HOME }}  # ERROR, $HOME is not defined (as a Jinja2 variable)!
+   {{ "$HOME" }}  # OK, prints the literal string "$HOME"
+   {{ environ["HOME"] }}  # OK, prints the value of $HOME
+
+Note however that this code executes during file parsing, at start-up, on the
+scheduler run host, because that is when template preprocessing is done. It
+does not read the environment at run time on the job host, even if the
+code appears in (i.e., writes to) a task definition section of the workflow.
+
+Similarly, any Jinja2 access (via Python) to the filesystem will happen during
+file parsing on the scheduler run host, not when jobs run on job hosts.
+
+
+Environment Variables and Shell Scripting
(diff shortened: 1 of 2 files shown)


src/user-guide/writing-workflows/configuration.rst
```diff
@@ -62,10 +62,137 @@ documentation, configuration files, etc. When the workflow is :ref:`installed
...
+
+Cylc workflow configurations are not executable scripts or programs. Rather,
+they configure, at start-up, the Cylc scheduler program to run a particular
+workflow.
+
+The scheduler understands a static syntax that contains only ``[section
+headings]`` and configuration items of the form ``platform = hpc1``. These
+configuration items are not even "variables" that can be referenced elsewhere
+in the file. Some values may look complicated, but they are still just text
+strings with content that is meaningful to the scheduler.
```
oliver-sanders, line 78: The tone here is a bit negative, Cylc provides a powerful configuration format, it just isn't dynamic.

Maybe just explain that the Cylc config is static, but can be augmented by template engines such as Jinja2 and explain that these engines are run *before* startup.

src/user-guide/writing-workflows/configuration.rst
```diff
@@ -62,10 +62,137 @@ documentation, configuration files, etc. When the workflow is :ref:`installed
...
+
+Nevertheless a `flow.cylc` file may appear to contain several kinds of embedded
+code:
+
+    - Python-like Jinja2 or EmPy code, such as ``{% set PLANET = "earth" %}``
+    - Bash shell variable assignments such as ``PLANET=${PLANET:-earth}``
+    - Bash shell scripting, such as ``script = "run-model.exe /path/to/data"``
```
oliver-sanders, line 85:
```suggestion
- Python-like Jinja2 or EmPy code, such as ``{% set PLANET = "earth" %}``
- Bash shell variable assignments such as ``PLANET=${PLANET:-earth}``
- Bash shell scripting, such as ``script = "run-model.exe /path/to/data"``
```

src/user-guide/writing-workflows/configuration.rst
```diff
@@ -62,10 +62,137 @@ documentation, configuration files, etc. When the workflow is :ref:`installed
...
+
+The documentation below explains exactly what these mean in the context of a
+workflow configuration, and when the code is evaluated.
+
+
+Jinja2 (or EmPy) Code
+^^^^^^^^^^^^^^^^^^^^^
```
oliver-sanders, line 92: Link to the Jinja2/Empty bits of the user guilde from somewhere in here.

src/user-guide/writing-workflows/configuration.rst
```diff
@@ -62,10 +62,137 @@ documentation, configuration files, etc. When the workflow is :ref:`installed
...
+programmatically generating the static configuration format required by the
+scheduler.
+
+The template processor manipulates the surrounding text (which to it is
+entirely arbitrary) and the result must be valid configuration syntax that
+will pass validation by Cylc.
+
+This is a preprocessing step. Jinja2 code does not execute as the workflow runs.
+
+Use ``cylc view -j`` to see the result, with all Jinja2 code "processed out".
```
oliver-sanders, line 105: I think the `--process` option is the more generic advice?

```suggestion
Use ``cylc view -p`` to see the result, with all Jinja2 code "processed out".
```

src/user-guide/writing-workflows/configuration.rst
```diff
@@ -62,10 +62,137 @@ documentation, configuration files, etc. When the workflow is :ref:`installed
...
+
+.. code-block:: cylc
+
+   # flow.cylc
+   {% set PLATFORM = "hpc1" %}
+   platform = {{ PLATFORM }}  # OK
+   {{ "Platform is: " ~ PLATFORM }} # OK
+ 
+Jinja2 code does not have access to the environment via normal shell syntax
+either, but you can read the local environment with native Jinja2 syntax:
```
oliver-sanders, line 127: This is duplicating the Jinja2 section somewhat.

More positive spin:

```suggestion
The scheduler's environment is available via the ``environ`` dictionary.
```

oliver-sanders (commented):
I fear the tone and length of this article makes Cylc seem limited and complex, which I don't think is right.

The main takeaway points are:

* Template processing happens at workflow startup.
* So you can't use Jinja2 to dynamically generate job scripts.
* Template engines don't have access to the worflow config, but you can provide information to it using variables.
* `cylc view` is a useful tool when working with template engines.

I would be tempted to convey these messages through a couple of notes or warnings in the Jinja2 section / tutorial.

hjoliver:
>  I fear the tone and length of this article makes Cylc seem limited and complex, which I don't think is right.

On re-reading it, I agree.  I'll try to simplify and re-tone it!

> The main takeaway points are: ... I would be tempted to convey these messages through a couple of notes or warnings in the Jinja2 section / tutorial.

It's worse than that though, and involves shell code as well as Jinja2. 

Take this recent forum post from an (experienced!) user:

https://cylc.discourse.group/t/how-do-you-use-a-variable-defined-in-environment-in-the-directives-section/760

and shortly after:

https://cylc.discourse.group/t/how-do-you-use-a-variable-defined-in-environment-in-the-directives-section/760/3

hjoliver:
OK thanks for the feedback. I've stripped that sucker back to the bare bones (I think). Try it now for size.

src/user-guide/writing-workflows/configuration.rst
```diff
@@ -62,10 +62,53 @@ documentation, configuration files, etc. When the workflow is :ref:`installed
...
+Understanding Code in Workflow Configurations
+---------------------------------------------
+
+A workflow configuration is not executable code. It configures the scheduler
+program to run your workflow. But a `flow.cylc` file may contain:
+
+    - Embedded Python-like Jinja2 or EmPy templating code, such as
+      ``{% set PLANET = "earth" %}``
+    - Bash shell variable assignments and scripting, such as
+      ``script = "run-model.exe /path/to/data"``
```
oliver-sanders, line 76:
```suggestion
program to run your workflow. A `flow.cylc` file may contain:

- Embedded Python-like Jinja2 or EmPy templating code, such as
  ``{% set PLANET = "earth" %}``
- Bash shell variable assignments and scripting, such as
  ``script = "run-model.exe /path/to/data"``
```

src/user-guide/writing-workflows/configuration.rst
```diff
@@ -62,10 +62,53 @@ documentation, configuration files, etc. When the workflow is :ref:`installed
...
+
+    - Jinja2 code can only reference Jinja2 variables, not Cylc config items.
+    - Jinja2 (like Python) has its own syntax for reading environment variables.
+    - Jinja2 code that reads the environment or the filesystem will do so
+      during config file parsing on the scheduler run host, not on job hosts.
+      Beware of doing this in task definitions - do you want the scheduler
+      environment to affect shell code that runs in the job environment?
+    - Shell code destined for the job script can read the environment or
+      access the filesystem as the job runs on the job host, not on the
+      scheduler host.
```
oliver-sanders, line 105:
```suggestion
- Jinja2 code is evaluated once when the workflow is started.
- Jinja2 code can only reference Jinja2 variables, not Cylc config items.
- Jinja2 (like Python) has its own syntax for reading environment variables.
- Jinja2 code that reads the environment or the filesystem will do so
  during config file parsing on the scheduler run host, not on job hosts.
  Beware of doing this in task definitions - do you want the scheduler
  environment to affect shell code that runs in the job environment?
- Shell code destined for the job script can read the environment or
  access the filesystem as the job runs on the job host, not on the
  scheduler host.
```

oliver-sanders (approved):
Cheers

hjoliver:
TODO: add cross-referencing.

PR #653, opened 2023-09-27
Fix.tutorial

wxtim:
Prep for course delivered October 2023.

Paired with changes in https://github.com/cylc/cylc-flow/pull/5744

- Clarify runtime tutorial
- Add a `cylc install` to the runtime tutorial.
- Fix broken Symlink
- Exempted `https://linux.die.net/.*` linkcheck tests. The domain seems to have implemented more stringent origin meta checks - [see comment on Element](https://matrix.to/#/!hwZqSYihGPuhDdIzIP:matrix.org/$169591864225544fzfBe:matrix.org?via=matrix.org)

**Requirements check-list**

IMO we should not merge these PR's until @joehickson and I have had a chance to skim the whole tutorial.

### DIFF
--- a/src/tutorial/runtime/runtime-configuration.rst
+++ b/src/tutorial/runtime/runtime-configuration.rst
@@ -312,7 +312,7 @@ Start, Stop, Restart
          cylc get-resources api-key
 
       Add the following lines to the ``[runtime]`` section of the
-      :cylc:conf:`flow.cylc` file replacing
+      :cylc:conf:`flow.cylc` file
 
       .. code-block:: cylc
 
@@ -327,8 +327,8 @@ Start, Stop, Restart
       Add three more ``get_observations`` tasks for each of the remaining
       weather stations.
 
-      You will need the API key and the site codes for the other three
-      weather stations, which are:
+      You will need to use the same API key for the other three
+      weather stations, and set the ``SITE_ID`` values, which are:
 
       * Camborne - ``3808``
       * Shetland - ``3005``
--- a/src/tutorial/runtime/runtime-configuration.rst
+++ b/src/tutorial/runtime/runtime-configuration.rst
@@ -311,8 +311,8 @@ Start, Stop, Restart
 
          cylc get-resources api-key
 
-      Add the following lines to the bottom of the :cylc:conf:`flow.cylc` file replacing
-      ``xxx...`` with your API key:
+      Add the following lines to the ``[runtime]`` section of the
+      :cylc:conf:`flow.cylc` file replacing
 
       .. code-block:: cylc
 
@@ -327,7 +327,8 @@ Start, Stop, Restart
       Add three more ``get_observations`` tasks for each of the remaining
       weather stations.
 
-      You will need the codes for the other three weather stations, which are:
+      You will need the API key and the site codes for the other three
+      weather stations, which are:
 
       * Camborne - ``3808``
       * Shetland - ``3005``
@@ -379,6 +380,7 @@ Start, Stop, Restart
 
       .. code-block:: bash
 
+         cylc install
          cylc play runtime-tutorial
 
       If all goes well the workflow will startup and the tasks will run and
--- a/src/tutorial/runtime/runtime-configuration.rst
+++ b/src/tutorial/runtime/runtime-configuration.rst
@@ -431,6 +431,7 @@ Start, Stop, Restart
       .. code-block:: bash
 
          cylc validate .
+         cylc install
 
       .. TODO: Add advice on what to do if the command fails.
 
--- a/src/user-guide/installing-workflows.rst
+++ b/src/user-guide/installing-workflows.rst
@@ -402,7 +402,8 @@ The run directory now looks as follows:
 
    If your workflow needs to create or install scripts or executables at runtime
    and you don't want Cylc to delete them on re-installation, you can use
-   ``bin`` and ``lib/python`` directories in the :ref:`workflow_share_directories`.
+   ``bin`` and ``lib/python`` directories in the
+   :ref:`workflow share directory<workflow_share_directories>`.
 
 
 Expected Errors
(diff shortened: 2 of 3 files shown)


src/tutorial/runtime/runtime-configuration.rst
```diff
@@ -311,8 +311,8 @@ Start, Stop, Restart
 
          cylc get-resources api-key
 
-      Add the following lines to the bottom of the :cylc:conf:`flow.cylc` file replacing
-      ``xxx...`` with your API key:
+      Add the following lines to the ``[runtime]`` section of the
+      :cylc:conf:`flow.cylc` file replacing
```
oliver-sanders, line 315: What was wrong with this instruction?

> flow.cylc file replacing ...

... replacing what???

src/tutorial/runtime/runtime-configuration.rst
```diff
@@ -327,7 +327,8 @@ Start, Stop, Restart
       Add three more ``get_observations`` tasks for each of the remaining
       weather stations.
 
-      You will need the codes for the other three weather stations, which are:
+      You will need the API key and the site codes for the other three
+      weather stations, which are:
```
oliver-sanders, line 331: What was wrong with this instruction?

src/user-guide/installing-workflows.rst
```diff
@@ -402,7 +402,7 @@ The run directory now looks as follows:
 
    If your workflow needs to create or install scripts or executables at runtime
    and you don't want Cylc to delete them on re-installation, you can use
-   ``bin`` and ``lib/python`` directories in the :ref:` workflow share directory <workflow_share_directories>`.
+   ``bin`` and ``lib/python`` directories in the :ref:`workflow_share_directories`.
```
oliver-sanders, line 405: ?

src/tutorial/runtime/runtime-configuration.rst
wxtim: As Joe observed we are modifying existing namespaces in the runtime section, not adding extra to the bottom of the config.

src/tutorial/runtime/runtime-configuration.rst
wxtim: > What was wrong with this instruction? I don't think this has changed since Cylc 7.

You need to set both `SITE_ID` _and_ `API_KEY`. IMO "codes" is not sufficiently clear that your are setting the same `API_KEY` and a different `SITE_ID`.  Although the context is such I'm happy to revert and ignore.

> This change makes it sound like they need a different API key for each site.

Agreed, modifying in line with this.

src/user-guide/installing-workflows.rst
wxtim: This link was broken. Dave spotted it whilst doing my review.

src/tutorial/runtime/runtime-configuration.rst
oliver-sanders: Hmm, I'm sure the way it was written was correct, has the tutorial changed since Cylc 7?

src/tutorial/runtime/runtime-configuration.rst
oliver-sanders: Or leave as is and add "replacing the API key as before".

src/user-guide/installing-workflows.rst
oliver-sanders: Eh? This change didn't change the link, it just removed the placeholder text "workflow share directory".

wxtim: Didn't appear to - have a look at it on the compiled docs...

src/tutorial/runtime/runtime-configuration.rst
wxtim: It was correct, but this is more correct.

src/user-guide/installing-workflows.rst
oliver-sanders: I'm not doubting the link was broken (I'm sure you tested it), but the change shouldn't have fixed it because the target is the same between the two examples.

Taking a look above, you can see there colour of the `:ref:` is different before and after so there must have been a syntax error, looks like it's the space:

```rst
:ref:` workflow share directory <workflow_share_directories>`
:ref:`workflow share directory <workflow_share_directories>`
```

So this should do it:

```suggestion
   ``bin`` and ``lib/python`` directories in the :ref:`workflow share directory <workflow_share_directories>`.
```

src/tutorial/runtime/runtime-configuration.rst
oliver-sanders: > has the tutorial changed since Cylc 7 like the other example in cylc-flow?

It has!

During the transfer of the tutorials some blank task sections appear to have been added:

https://github.com/metomi/rose/blob/2019.01.x/etc/tutorial/runtime-tutorial/suite.rc
https://github.com/cylc/cylc-flow/blob/master/cylc/flow/etc/tutorial/runtime-tutorial/flow.cylc

wxtim: OK, will change on the docs side

src/user-guide/installing-workflows.rst
wxtim: But the placeholder test is unecessary

oliver-sanders: Yes and no.

Your diff would change the text like so:

```diff
- ``bin`` and ``lib/python`` directories in the workflow share directory
+ ``bin`` and ``lib/python`` directories in the Workflow Share Directory
```

If you don't specify the placeholder text, then it defaults to the title of the section. This is a useful default, but there are two issues:

1. It can lead to problems when sections are renamed.
2. The capitalisation is typically unsuitable.

wxtim: > Someone took the time to add a placeholder here, we should preserve it.

Me, about a fortnight ago

wxtim:
```suggestion
   ``bin`` and ``lib/python`` directories in the :ref:`workflow share directory <workflow_share_directories>`.
```

oliver-sanders (approved):

joeHickson:
configuration-consolidation/jinja2.html 
#3 - #!Jinja2 shebang already present at the top of the file

markgrahamdawson (approved):

PR #660, opened 2023-10-04
clarify runahead limit

oliver-sanders:
The "consecutive" terminology can be a little confusing.

See also: https://github.com/cylc/cylc-flow/pull/5754

**Requirements check-list**


src/user-guide/writing-workflows/scheduling.rst
```diff
@@ -2015,22 +2015,25 @@ tools to limit workflow activity.
...
+The runahead limit can be configured for a workflow by
+:cylc:conf:`[scheduling]runahead limit`. It defines the span of cycles which
+are permitted to run as measured from the oldest active cycle point. The
+runahead limit can be defined in two ways, as an integer or datetime span.
+
+.. rubric:: Integer Format
 
 The integer format ``Pn`` is an interval that spans ``n+1`` consecutive cycle
 points, regardless of how the cycle point values increment from one point to
 the next (i.e. regardless of the cycling interval).
```
dpmatthews, line 2033: I'm not really sure what "regardless of the cycling interval" means here.
Maybe just say:
"The integer format ``Pn`` is an interval that spans ``n+1`` consecutive cycle points as defined by the graph"

src/user-guide/writing-workflows/scheduling.rst
```diff
@@ -2015,22 +2015,25 @@ tools to limit workflow activity.
...
 
 For example, the default runahead limit of ``P4`` allows 5 active cycle points
```
dpmatthews, line 2035: I think "up to 5" would be better

src/user-guide/writing-workflows/scheduling.rst
```diff
@@ -2015,22 +2015,25 @@ tools to limit workflow activity.
...
-(base point plus the 4 next points) whether the cycling is defined by
-``R/^/P1`` or ``R/^/P3`` etc. (for integer cycling) or ``R/^/PT6H`` etc. (for
-datetime cycling).
+(the oldest active cycle plus the 4 next cycles).
 
 In the following example, tasks ``1/foo`` through ``4/foo`` will submit to
 run immediately at start-up, but ``5/foo`` will be held back until ``1/foo``
```
dpmatthews, line 2039: No! The example uses P2 at the moment. Either correct these or change the example to use P1?

dpmatthews (changes requested):

src/user-guide/writing-workflows/scheduling.rst
hjoliver: > I'm not really sure what "regardless of the cycling interval" means here

Th intention was, I think, to emphasize that a Pn limit determines the number of cycle points regardless of how big or small the interval between cycles is. (As compared to a duration limit, where the number of cycles will depend on the interval).

src/user-guide/writing-workflows/scheduling.rst
```diff
@@ -2015,22 +2015,25 @@ tools to limit workflow activity.
...
-be active at once. The base point of the runahead calculation is the current
-lowest-valued point with :term:`active` or :term:`incomplete` tasks present.
+The runahead limit prevents a workflow from getting too far ahead of the oldest
+active cycle point.
 
-The runahead limit can be set with :cylc:conf:`[scheduling]runahead limit`,
-to an integer interval (for datetime or integer cycling workflows) or to
-a datetime interval (datetime cycling only).
+A cycle is considered to be active if it contains any :term:`active` or
+:term:`incomplete` tasks.
```
hjoliver, line 2022: We need to get our story straight on exactly what "active cycle points" (aka the "active window" of the workflow) contain(s):
- literally active tasks (submitted, running)
- incomplete tasks (finished without completing required outputs)
- waiting tasks that are ready to run as far as the task graph is concerned, but are still waiting on
  - clock triggers
  - xtriggers
  - queue release
  - or hold

This isn't the right place to say all of that, but we should put it somewhere and link all references to "active cycle points" to it.

src/user-guide/writing-workflows/scheduling.rst
```diff
@@ -2015,22 +2015,25 @@ tools to limit workflow activity.
...
+
+The runahead limit can be configured for a workflow by
+:cylc:conf:`[scheduling]runahead limit`. It defines the span of cycles which
+are permitted to run as measured from the oldest active cycle point. The
+runahead limit can be defined in two ways, as an integer or datetime span.
```
hjoliver, line 2027:
```suggestion
:cylc:conf:`[scheduling]runahead limit`. It can be defined in two ways, as an integer or datetime interval.
```
That repeats the information just above.

hjoliver (changes requested):

src/user-guide/writing-workflows/scheduling.rst
dpmatthews: What about tasks with partially satisfied dependencies?
I assume an active cycle is one with a task in the n=0 window?

hjoliver: Uh, yes! Especially after the "Cylc set" PR (which ditches the hidden pool)

oliver-sanders: Ok, it's sounding like we need a new glossary entry to describe the meaning of "active cycle".

The definition should include:
* Active tasks.
* Incomplete tasks.
* Waiting tasks that have one or more of their dependencies satisfied.
* Waiting parentless tasks.

Good?

dpmatthews: I think we should define an active cycle as one which contains a task in the active ("n=0") window.
This term is already defined in the glossary but is wrong - I think we should update it to include:
* Active tasks
* Incomplete tasks
* Active-waiting tasks (we might need to tweak the existing definition in the glossary for these)
* Waiting tasks with some, but not all, of their pre-requisites met

> Waiting parentless tasks

I don't understand the relevance of parentless here.

hjoliver: Plus (as I said above):
- waiting tasks that are ready to run as far as the task graph is concerned, but are still waiting on: queues, hold, clock and xtriggers

> >    Waiting parentless tasks
> 
> I don't understand the relevance of parentless here.

Parentless tasks get auto-spawned to the runahead limit (although, after @dwsutherland 's extant PR clock-triggered ones won't). That is technically covered by "ready to run as far at the task graph is concerned" but we should document it anyway - might not be obvious to users.

src/user-guide/running-workflows/index.rst
```diff
@@ -7,7 +7,6 @@ Running Workflows
    :maxdepth: 2
 
    scheduler-start-up
-   task-job-states
    tasks-jobs-ui
```
oliver-sanders, line 10: Merged the two short pages on tasks in the UI.

src/user-guide/running-workflows/tasks-jobs-ui.rst
```diff
@@ -1,45 +1,153 @@
+.. _task-job-states:
```
oliver-sanders, line 1: Pre-existing content moved in from `task-job-states.rst`.

hjoliver (approved):
Great.

hjoliver:
Needed for 8.3.0

PR #671, opened 2023-11-22
Add documentation for xtriggers entry_point

ColemanTom:
**Requirements check-list**

I've attempted to add documentation for new functionality provided by https://github.com/cylc/cylc-flow/pull/5831. I hope it works properly. I think there should be a link between the external triggers documentation and the plugin documentation, but I'm not sure how to do that. Very specific advice would be appreciated.

When I tried to build this as myself I had issues and have not had time to try and debug, so I'm not sure how it renders (or if it is even valid).

### DIFF
--- a/src/plugins/xtriggers/index.rst
+++ b/src/plugins/xtriggers/index.rst
@@ -46,7 +46,7 @@ entry points. Each xtrigger is registered individually.
    :caption: ``pyproject.toml``
 
    [project.entry-points."cylc.xtriggers"]
-   foo = my_package.foo:foo
-   bar = my_package.foo:bar
-   baz = my_package.baz:baz
+   foo = "my_package.foo:foo"
+   bar = "my_package.foo:bar"
+   baz = "my_package.baz:baz"
 
--- a/src/plugins/xtriggers/index.rst
+++ b/src/plugins/xtriggers/index.rst
@@ -1,0 +1,52 @@
+Xtrigger Plugins
+======================================
+
+Xtrigger plugins allow you to install and use xtriggers without them being
+in your ``CYLC_PYTHONPATH``.
+
+
+Built In Plugins
+----------------
+
+Cylc Flow provides the following xtriggers.
+
+.. autosummary::
+   :toctree: built-in
+   :template: docstring_only.rst
+
+   cylc.flow.xtriggers.echo
+   cylc.flow.xtriggers.workflow_state
+   cylc.flow.xtriggers.xrandom
+
+.. Note: Autosummary generates files in this directory, these are cleaned
+         up by `make clean`.
+
+Developing ``xtrigger`` plugins
+-------------------------------
+
+Cylc uses entry points registered by setuptools to search for xtrigger
+plugins.
+
+Example
+^^^^^^^
+
+Plugins are registered by registering them with the ``cylc.xtriggers``
+entry points. Each xtrigger is registered individually.
+
+.. code-block:: ini
+   :caption: ``setup.cfg``
+
+   [options.entry_points]
+       cylc.xtriggers =
+           foo = my_package.foo:foo
+           bar = my_package.foo:bar
+           baz = my_package.baz:baz
+
+.. code-block:: toml
+   :caption: ``pyproject.toml``
+
+   [project.entry-points."cylc.xtriggers"]
+   foo = my_package.foo:foo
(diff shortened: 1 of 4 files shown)


src/plugins/xtriggers/index.rst
```diff
@@ -0,0 +1,74 @@
...
+
+Plugins are registered by registering them with the ``cylc.xtriggers``
+entry points:
+
+.. code-block:: toml
+   :caption: ``setup.cfg``
+
+   [options.entry_points]
+       cylc.xtriggers =
+           my_package = my_package.entry_install:xtriggers
```
oliver-sanders, line 58: I'm not keen on using a function to load xtriggers from a package. 

I would recommend registering each xtrigger individually like this:

```diff
    [options.entry_points]
        cylc.xtriggers =
-           my_package = my_package.entry_install:xtriggers
+           foo = my_package.foo:foo
+           bar = my_package.bar:bar
+           baz = my_package.baz:baz
```

The reason for this is that it follows the pattern of existing plugins and reduces the import overheads, i.e. if the baz xtrigger is used in a workflow, only the baz module (and any dependencies) need be imported, whereas loading them from a function, all modules (and dependencies) get imported if any are used.

ColemanTom: Thanks. I'm learning a lot from this one. I agree your approach is much better and is adopted.

src/plugins/xtriggers/index.rst
```diff
@@ -0,0 +1,53 @@
...
+----------------
+
+Cylc Flow provides the following xtriggers.
+
+.. autosummary::
+   :toctree: built-in
+   :template: docstring_only.rst
+
+   cylc.flow.xtriggers.echo
+   cylc.flow.xtriggers.wall_clock
```
MetRonnie, line 18: Unfortunately the fly in the ointment here is that the `wall_clock` function has a different signature in the code compared to real usage. Not sure what the reason for this was, but it was done in https://github.com/cylc/cylc-flow/pull/4511. Maybe @hjoliver will be able to shed some light

ColemanTom: I don't really understand this, outside my area of knowledge. Do I need to change anything or await advice from @hjoliver ?

src/plugins/xtriggers/index.rst
```diff
@@ -0,0 +1,53 @@
...
+           bar = my_package.foo:bar
+           baz = my_package.baz:baz
+
+.. code-block:: toml
+   :caption: ``pyproject.toml``
+
+   [project.entry-points."cylc.xtriggers"]
+   foo = my_package.foo:foo
+   bar = my_package.foo:bar
+   baz = my_package.baz:baz
```
MetRonnie, line 52: These need to be strings in TOML
```suggestion
   foo = "my_package.foo:foo"
   bar = "my_package.foo:bar"
   baz = "my_package.baz:baz"
```

src/plugins/xtriggers/index.rst
oliver-sanders: The `wall_clock` xtrigger isn't really an xtrigger because it would be inefficient to call this function in a subprocess over and over. So instead it's actually run synchronously within the main code, the trigger time is computed from the `offset` and the cycle point in advance to avoid excess computation.

This sort of problem will go away with https://github.com/cylc/cylc-flow/issues/3497 as this will avoid polling a function allowing it to maintain local state.

Until then, maybe just subtract the `wall_clock` function from the autodocumented reference.

ColemanTom: Thanks, removed wall_clock. I also noticed I had PYTHONPATH instead of CYLC_PYTHONPATH, so I updated that too.

ColemanTom:
test failure is casued by openbsd website being down. Nothing I can do to fix that. It could change to the freebsd version of the `ssh_config` documentation, which may be a more reliable website given its customers?

MetRonnie:
Could you add `src/plugins/xtriggers/built-in` to `.gitignore` please

ColemanTom:
> Could you add `src/plugins/xtriggers/built-in` to `.gitignore` please (autodoc generates files into this directory)

Done

MetRonnie (approved):

MetRonnie:
@oliver-sanders Are you happy with my single review or did you want to re-review?

Issue #106, opened 2019-12-02
document built-in xtriggers


oliver-sanders:
Follow-on from https://github.com/cylc/cylc-flow/pull/3457
Attempt after: https://github.com/cylc/cylc-flow/issues/3456

Auto-document the Cylc Flow built-in Xtriggers from their doc strings.

hjoliver:
Also, we need to better document how users call built-in xtrigger functions.  REF https://cylc.discourse.group/t/xtrigger-suite-state-function-usage/219

PR #673, opened 2023-11-22
Autodoc built-in xtrigger functions

MetRonnie:
The `debug` arg is not around anymore. By autofunction'ing, these will always be valid. The downside is pulling in the docstring is quite verbose...

**Requirements check-list**

### DIFF
--- a/src/conf.py
+++ b/src/conf.py
@@ -109,6 +109,10 @@ autosummary_generate = True
 autosummary_generate_overwrite = True
 autosummary_imported_members = False
 
+# Autodoc type hints can look very messy if they are included in the signature,
+# so we only include them in the description instead:
+autodoc_typehints = 'description'
+
 # Mapping to other Sphinx projects we want to import references from.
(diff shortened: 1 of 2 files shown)


oliver-sanders:
Note, this PR will conflict with https://github.com/cylc/cylc-doc/pull/671 which also auto-documents xtriggers, but in the plugins section (the xtriggers section is more appropriate).

We could probably do with pulling any relevant text out of cylc-doc and into the xtrigger docstrings to keep it all together?

For maintainability, I think we should probably go along the same lines at the job runner handlers and put the `autosummary` or `autodoc` directive in the cylc-flow code, then autodocument the __init__ file, WDYT:

https://github.com/cylc/cylc-flow/blob/master/cylc/flow/job_runner_handlers/__init__.py

oliver-sanders:
Closes https://github.com/cylc/cylc-doc/issues/106

MetRonnie:
Superseded by #686

PR #682, opened 2024-01-30
Add words to spellcheck dictionary

MetRonnie:
Autodocumented exceptions introduced in #680 causing spellcheck to fail

**Requirements check-list**


oliver-sanders:
I'm not sure how sustainable spellchecking is going to be if it's going to complain over things like vairable names (kwargs). Not sure what the long-term solution is here.

wxtim (approved):

PR #683, opened 2024-01-30
add reference labels

oliver-sanders:
Add a couple of labels so we can reference these sections from other places.

**Requirements check-list**

### DIFF
--- a/src/user-guide/writing-workflows/scheduling.rst
+++ b/src/user-guide/writing-workflows/scheduling.rst
@@ -332,6 +332,8 @@ There are three ISO 8601 recurrence formats supported by Cylc, detailed below
 in order from most commonly used to least commonly used.
 
 
+.. _user_guide.cycling_format_3:
+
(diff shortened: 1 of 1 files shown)


oliver-sanders:
Test failures are unrelated.

They are the result of spellchecking running over the API docs and complaining about variable names like `kwargs` and `cfg`.

As we expand our API docs, spellchecking is going to be increasingly difficult to manage.

wxtim:
> Test failures are unrelated. Fixed on #682
> 
> They are the result of spellchecking running over the API docs and complaining about variable names like `kwargs` and `cfg`.
> 
> As we expand our API docs, spellchecking is going to be increasingly difficult to manage.

Yes. Sadly.

wxtim (approved):

oliver-sanders:
> Although I think that it's worth considering whether non dictionary words are reasonable in code

Yes they are!

PR #685, opened 2024-02-01
improve workflow events documentation

### LINKED ISSUE #684: Document meaning of inactivity timeout
The inactivity timeout is not affected by the presence of unsatisfied xtriggers (including clock triggers), which might be surprising if you abort on inactivity timeout in a clock-triggered workflow.

It seems a bit aggressive to abort on "inactivity" if there are active tasks and/or active xtriggers present.

On the other hand, if we made it more friendly, it might just be equivalent to the stall timeout.

I think this feature was originally added primarily for use in functional tests, where we might assume something has gone wrong if nothing is happening, even if the workflow is not technically stalled.

At the very least, we should document for users exactly what "inactivity" means.

oliver-sanders:
* Closes #684

**Requirements check-list**

### DIFF
--- a/src/user-guide/writing-workflows/scheduler.rst
+++ b/src/user-guide/writing-workflows/scheduler.rst
@@ -15,9 +15,12 @@ workflows.
 Workflow Events
 ---------------
 
-Cylc supports workflow events which pertain to the :term:`scheduler` e.g.
-``startup`` and ``shutdown`` and task events e.g. ``submitted`` and ``failed``
-which pertain to :term:`tasks <task>`. This section covers workflow events, for
+There are two types of event in Cylc:
+
+* workflow events e.g. ``startup`` and ``shutdown``, which pertain to the :term:`scheduler` 
+* task events e.g. ``submitted`` and ``failed``, which pertain to :term:`tasks <task>`.
+
+This section covers workflow events, for
 task events see :ref:`user_guide.runtime.task_event_handling`.
 
 .. rubric:: Event Handlers
@@ -75,6 +78,8 @@ Some workflow events have related configurations e.g. for setting the timeout.
 
    The timer starts counting down at scheduler startup. It resets on workflow
    restart.
+
+   Note, the `abort` event is not related to "Abort On Event" handlers.
 
 .. describe:: stall
 
--- a/src/user-guide/writing-workflows/scheduler.rst
+++ b/src/user-guide/writing-workflows/scheduler.rst
@@ -10,53 +10,110 @@ workflows.
 
 
 .. _user_guide.scheduler.workflow_event_handling:
+.. _user_guide.scheduler.workflow_events:
 
-Workflow Event Handling
------------------------
+Workflow Events
+---------------
 
-Workflow event handlers allow configured commands to run when workflow events
-occur.
+Cylc supports workflow events which pertain to the :term:`scheduler` e.g.
+``startup`` and ``shutdown`` and task events e.g. ``submitted`` and ``failed``
+which pertain to :term:`tasks <task>`. This section covers workflow events, for
+task events see :ref:`user_guide.runtime.task_event_handling`.
 
-.. note::
+.. rubric:: Event Handlers
 
-   Cylc supports workflow events e.g. ``startup`` and ``shutdown``
-   and task events e.g. ``submitted`` and ``failed``.
-
-   See also :ref:`user_guide.runtime.task_event_handling`.
-
-Workflow event handlers are configured by:
+Workflow events have "handlers" (i.e. hooks) which allow configured commands to
+run when workflow events occur. These can be configured by:
 
 * :cylc:conf:`flow.cylc[scheduler][events]` (per workflow)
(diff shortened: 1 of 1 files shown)


src/user-guide/writing-workflows/scheduler.rst
```diff
@@ -10,53 +10,110 @@ workflows.
...
-.. note::
+Workflow Events
+---------------
 
-   Cylc supports workflow events e.g. ``startup`` and ``shutdown``
-   and task events e.g. ``submitted`` and ``failed``.
+Cylc supports workflow events which pertain to the :term:`scheduler` e.g.
+``startup`` and ``shutdown`` and task events e.g. ``submitted`` and ``failed``
+which pertain to :term:`tasks <task>`. This section covers workflow events, for
+task events see :ref:`user_guide.runtime.task_event_handling`.
```
MetRonnie, line 21:
```suggestion
There are two types of event in Cylc:
* workflow events e.g. ``startup`` and ``shutdown``, which pertain to the :term:`scheduler` 
* task events e.g. ``submitted`` and ``failed``, which pertain to :term:`tasks <task>`.

This section covers workflow events, for
task events see :ref:`user_guide.runtime.task_event_handling`.
```

src/user-guide/writing-workflows/scheduler.rst
```diff
@@ -10,53 +10,110 @@ workflows.
...
+
+   E.G. using one of these commands ``cylc play``, ``cylc vip`` or ``cylc vr``.
+
+.. describe:: shutdown
+
+   :Event Handler: `shutdown handlers`
+
+   The scheduler was shut down.
+
+   E.G. using the ``cylc stop`` command.
```
MetRonnie, line 59: Doesn't this also apply to aborts and the workflow reaching completion naturally?

oliver-sanders: Natural shutdown yes, not sure about aborts, I would assume not (not a real shutdown, would be odd to get both an aborted and a shutdown event).

MetRonnie: I have tested with
```
[scheduler]
    [[events]]
        abort on workflow timeout = True
        workflow timeout = PT1S
        abort handlers = echo "ALPHA"
        shutdown handlers = echo "SIERRA"
```

Only the shutdown handler seemed to run
```
ERROR - Workflow shutting down - "abort on workflow timeout" is set
INFO - [('workflow-event-handler-00', 'shutdown') cmd] echo "SIERRA" 'shutdown' 'wflow1' '"abort on workflow timeout" is set'
    [('workflow-event-handler-00', 'shutdown') ret_code] 0
    [('workflow-event-handler-00', 'shutdown') out] SIERRA shutdown wflow1 "abort on workflow timeout" is set
INFO - DONE
```

oliver-sanders: Not sure that's a bug. There's a specific event for workflow timeout.

oliver-sanders: From testing:

* The `abort` handler is only for uncaught internal errors and not for `abort on ...` configs. The language is leading but this behaviour makes sense as we have already fired a more specific event for these cases.
* The `shutdown` handler gets called for pretty much all cases.

src/user-guide/writing-workflows/scheduler.rst
```diff
@@ -10,53 +10,110 @@ workflows.
...
-startup
-   The scheduler started running the workflow.
-shutdown
-   The workflow finished and the scheduler will shut down.
-abort
    The scheduler shut down early with error status, due to a fatal error
    condition or a configured timeout.
-workflow timeout
+
+.. describe:: workflow timeout
```
oliver-sanders, line 68:
```suggestion
   condition or a configured timeout.
   
   Note, the `abort` event is not related to "Abort On Event" handlers.

.. describe:: workflow timeout
```

oliver-sanders: @metronnie, resolution to https://github.com/cylc/cylc-doc/pull/685/files#r1478662197

ok?

MetRonnie: How about "not raised by" instead of "not related to"? But yes that's ok

MetRonnie: Still need to apply this suggestion

oliver-sanders: Ach, it won't apply, will do manually...

src/user-guide/writing-workflows/scheduler.rst
```diff
@@ -10,53 +10,115 @@ workflows.
...
+
    The workflow run timed out.
-stall
-   The workflow stalled.
-stall timeout
+
+   The timer starts counting down at scheduler startup. It resets on workflow
+   restart.
+
+   Note, the `abort` event is not related to "Abort On Event" handlers.
```
MetRonnie, line 82:
```suggestion
   Note, the ``abort`` event is not raised by "Abort On Event" handlers.
```

MetRonnie (approved):

Issue #626, opened 2023-07-17
interventions


oliver-sanders:
A new section of the user-guide covering common interventions.

E.G:

* How to "ignore" a task failure (i.e. set task to succeeded).
* How to "continue" a completed workflow (#525).
* How to "re-run" a workflow (i.e. `--flow=new`)

Some of these interventions are covered here: https://github.com/cylc/cylc-admin/blob/master/docs/proposal-interventions.md

These issues will change the documented interventions:

* https://github.com/cylc/cylc-flow/issues/5643
* https://github.com/cylc/cylc-flow/issues/5642
* https://github.com/cylc/cylc-flow/issues/5641
* https://github.com/cylc/cylc-flow/issues/5640

Others will need a bit of research / head-scratching.

Ideally these should be tested somehow, e.g. using a reftest.

oliver-sanders:
It would be a lot easier to explain these interventions with GUI videos, however, that is going to be tricky.

Somewhat related to e2e functional testing of the UI with live data and real workflows which would be required to truly test these interventions (and possibly record videos for the docs).

PR #723, opened 2024-04-26
user-guide: add interventions section

oliver-sanders:
* Closes https://github.com/cylc/cylc-doc/issues/626
* Attempt to document the most common operator interventions.
* Add the sphinx-design library to the docs environment for tabbed selectors.

Once the skip-mode and remove features are implemented, we should extend this.

**Requirements check-list**


src/user-guide/interventions/index.rst
```diff
@@ -0,0 +1,433 @@
...
+
+.. tab-set::
+
+   .. tab-item:: GUI
+      :sync: gui
+
+      The Cylc web app (GUI) can be launched with the ``cylc gui`` command.
+
+      Some sites provide a central deployment for you where you log in via the
+      web browser rather than launching it youself.
```
oliver-sanders, line 23:
```suggestion
      web browser rather than launching it yourself.
```

src/user-guide/interventions/index.rst
```diff
@@ -0,0 +1,433 @@
...
+
+   .. tab-item:: With the Tui
+      :sync: tui
+
+      Cylc Tui is an interactive in-terminal application, kinda like a minimal
```
wxtim, line 28: Excessive informality?
```suggestion
      Cylc Tui is an interactive in-terminal application, like a minimal
```

wxtim (approved):
1 Comment, nothing serious. I think that this is a very useful and powerful page. 👍🏼

src/user-guide/interventions/index.rst
```diff
@@ -0,0 +1,433 @@
...
+         $ cylc set <workflow>//<task>
+
+      By default, this sets the "succeeded" output. Use the ``--output`` option
+      to specify a different output.
+
+You can also :ref:`set task prerequisites <interventions.set-task-prerequisites>`.
+Should I set task outputs or prerequisites?
+
+* If you set a task's outputs, then tasks downstream of it may start to run.
+* If you set a task's prerequisites, this will not happen.
```
hjoliver, line 192:
```suggestion
* If you set a task's prerequisites, the task itself may start to run.
```
(In both cases, we could perhaps say exactly what "may" means, but I quite like this minimalist description).

src/user-guide/interventions/index.rst
```diff
@@ -0,0 +1,433 @@
...
+
+         $ cylc set \
+         >   --pre <prereq-cycle>/<prereq-task>:<prereq-output> \
+         >   <workflow>//<cycle><task>
+
+You can also :ref:`set task outputs <interventions.set-task-outputs>`.
+Should I set task outputs or prerequisites?
+
+* If you set a task's outputs, then tasks downstream of it may start to run.
+* If you set a task's prerequisites, this will not happen.
```
hjoliver, line 257:
```suggestion
* If you set a task's prerequisites, the task itself may start to run.
```

src/user-guide/interventions/index.rst
```diff
@@ -0,0 +1,433 @@
...
+   [scheduling]
+       cycling mode = integer
+       initial cycle point = 1
+       runahead limit = P1
+       [[graph]]
+           P1 = """
+               start => switch
+               switch:normal? => normal
+               switch:alternate? => alternate
+               normal | alternate => end
```
hjoliver, line 274: This has the problem that @MetRonnie and I flagged on the optional-outputs-extension PR. It looks nice, but novice users might not understand that the `normal` thing in `switch:normal` is not the same thing as in `=> normal`.  

Not critical though, can punt it to https://github.com/cylc/cylc-flow/issues/6085

src/user-guide/interventions/index.rst
```diff
@@ -0,0 +1,433 @@
...
+
+   .. tab-item:: GUI
+      :sync: gui
+
+      .. image:: edit-a-tasks-configuration.gui.gif
+         :width: 75%
+
+      .. note::
+
+         Any changes you make, apply only to this one instance of the task, not
```
hjoliver, line 357:
```suggestion
         Any changes you make apply only to this one instance of the task, not
```

src/user-guide/interventions/index.rst
```diff
@@ -0,0 +1,433 @@
...
+|
+
+.. _interventions.edit-the-workflow-configuration:
+
+Edit The Workflow Configuration
+-------------------------------
+
+:Example:
+   I want to change the configuration of multiple tasks or add/remove tasks
+   without stopping the workflow or starting it from the beginning again.
```
hjoliver, line 382: This mentions adding or removing tasks from the graph, which has some caveats, so I think we should link here to the "Restarting or reloading after graph changes" section.

hjoliver (approved):
Excellent!  Just a few minor suggestions.

src/user-guide/interventions/index.rst
wxtim: The question this statement would raise in my mind is:

Does that mean that if you set a tasks prerequisites it may run, but will not automatically trigger downstream tasks?

That's the question I would expect a user reading this to ask, both with and without Hilary's edits.

src/user-guide/interventions/index.rst
oliver-sanders: IMO, it's clearer this way than the alternative but open to reasonable suggestions.

src/user-guide/interventions/index.rst
oliver-sanders: Added a note.

src/user-guide/interventions/index.rst
hjoliver: I think it's reasonable as-is (with my edits), because downstream tasks won't run immediately in the set-prerequisite case. The target task has to run first and generate the outputs, and even that may not happen immediately (might be other prerequisites).

src/user-guide/interventions/index.rst
hjoliver: No need to change it on this PR anyway.

oliver-sanders:
Build error should be fixed by: https://github.com/cylc/cylc-flow/pull/6087

hjoliver:
(local build successful)

PR #727, opened 2024-05-01
Some doc updates for 8.3.0.dev changes.

hjoliver:
A quick blast through the docs (mostly the glossary) whilst trying not to overlap too much with the optional-outputs-extension, which should be documented separately (mostly by pasting from the proposal).

(I did add a glossary item on output completion, needed for cross-referencing).

### note on "incomplete"

We agreed not to define an "incomplete task" as **a finished task with incomplete outputs** (because that made "incomplete" different to "not complete" ). 

Now, "completion" is tied entirely to "output completion". That means it is legit to say that a task is incomplete if its outputs are incomplete, whether or not it has finished. That's convenient and I've used it a couple of times here. E.g. it means we can say a stall occurs if there is nothing left to run but *there are any incomplete tasks in the n=0 window" - that covers both finished-with-incomplete-outputs and waiting-with-partially-satisfied-prerequisites. Both of those are "active tasks" (i.e., in n=0) and have incomplete outputs.

**Requirements check-list**

### DIFF
--- a/src/7-to-8/major-changes/compatibility-mode.rst
+++ b/src/7-to-8/major-changes/compatibility-mode.rst
@@ -34,8 +34,9 @@ The ``suite.rc`` filename triggers a backward compatibility mode in which:
     branching)
 
 - ``succeeded`` task outputs are :ref:`*required* <User Guide Required Outputs>`,
-  so the scheduler will retain all other tasks with :term:`final status`
-  in the :term:`n=0 window <n-window>`.
+  so in the absence of suicide triggers the scheduler will retain other
+  :term:`final status` tasks in the :term:`n=0 window <n-window>` to stall the
+   workflow.
 
   - (in Cylc 8, **all** outputs are *required* unless marked as
     :ref:`*optional* <User Guide Optional Outputs>` by the new ``?`` syntax)
--- a/src/7-to-8/major-changes/compatibility-mode.rst
+++ b/src/7-to-8/major-changes/compatibility-mode.rst
@@ -36,7 +36,7 @@ The ``suite.rc`` filename triggers a backward compatibility mode in which:
 - ``succeeded`` task outputs are :ref:`*required* <User Guide Required Outputs>`,
   so in the absence of suicide triggers the scheduler will retain other
   :term:`final status` tasks in the :term:`n=0 window <n-window>` to stall the
-   workflow.
+  workflow.
 
   - (in Cylc 8, **all** outputs are *required* unless marked as
     :ref:`*optional* <User Guide Optional Outputs>` by the new ``?`` syntax)
--- a/src/7-to-8/major-changes/compatibility-mode.rst
+++ b/src/7-to-8/major-changes/compatibility-mode.rst
@@ -33,7 +33,7 @@ The ``suite.rc`` filename triggers a backward compatibility mode in which:
   - (Cylc 8 spawns tasks on demand, and suicide triggers are not needed for
     branching)
 
-- ``succeeded`` task outputs are :ref:`*required* <User Guide Required Outputs>`,
+- ``succeeded`` task outputs are :ref:`required <User Guide Required Outputs>`,
   so in the absence of suicide triggers the scheduler will retain other
   :term:`final status` tasks in the :term:`n=0 window <n-window>` to stall the
   workflow.
--- a/src/7-to-8/major-changes/compatibility-mode.rst
+++ b/src/7-to-8/major-changes/compatibility-mode.rst
@@ -34,7 +34,8 @@ The ``suite.rc`` filename triggers a backward compatibility mode in which:
     branching)
 
 - only ``succeeded`` task outputs are :ref:`*required* <User Guide Required Outputs>`,
-  meaning the scheduler will retain tasks that do not succeed as incomplete
+  so the scheduler will retain all non-succeeded :term:`finished tasks <finished task>`
+  in the :term:`n=0 window <n-window>`.
 
   - (in Cylc 8, **all** outputs are *required* unless marked as
     :ref:`*optional* <User Guide Optional Outputs>` by the new ``?`` syntax)
--- a/src/7-to-8/major-changes/compatibility-mode.rst
+++ b/src/7-to-8/major-changes/compatibility-mode.rst
@@ -33,8 +33,8 @@ The ``suite.rc`` filename triggers a backward compatibility mode in which:
   - (Cylc 8 spawns tasks on demand, and suicide triggers are not needed for
     branching)
 
-- only ``succeeded`` task outputs are :ref:`*required* <User Guide Required Outputs>`,
-  so the scheduler will retain all non-succeeded :term:`finished tasks <finished task>`
+- ``succeeded`` task outputs are :ref:`*required* <User Guide Required Outputs>`,
+  so the scheduler will retain all other tasks with :term:`final status`
   in the :term:`n=0 window <n-window>`.
 
   - (in Cylc 8, **all** outputs are *required* unless marked as
--- a/src/glossary.rst
+++ b/src/glossary.rst
@@ -106,11 +106,11 @@ Glossary
    window
    n-window
    active window
-      The UI provides a :term:`graph`-based window or view of the workflow at
+      The GUI provides a :term:`graph`-based window or view of the workflow at
       runtime, including tasks out to ``n`` graph edges from current
       :term:`active tasks <active task>`.
 
-      Active tasks form the ``n=0`` core of the :term:`n-window`.
+      Active tasks form the ``n=0``` window.
 
       .. seealso::
 
@@ -1383,17 +1383,6 @@ Glossary
       - ``:succeeded``, or ``:failed``
 
 
-   prerequisite
-   task prerequisite
-      A task's prerequisites are the :term:`outputs <task output>` of
-      parent tasks upstream in the graph that must be completed before
-      it can run. Tasks are said to "depend on" their prerequisites, hence
-      the term :term:`dependency graph <graph>`.
-
-      Tasks can depend on :term:`external triggers <xtrigger>` as well as
-      the outputs of other, upstream, tasks.
-
-
    output
    task output
       Task outputs mark the progression of a :term:`task` from waiting (for
@@ -1426,8 +1415,9 @@ Glossary
       A task's outputs are *complete* if its *output completion condition* 
       is satisfied.
 
-      The completion condition can be defined by the user, or otherwise
-      it is automatically generated according to the rules:
+      The completion condition can be defined by the user in
+      :cylc:conf:`flow.cylc[runtime][<namespace>]completion`,
+      otherwise it is automatically generated according to the rules:
 
       - All :term:`required outputs <required output>` (referenced in the graph)
         must be completed.
@@ -1444,10 +1434,13 @@ Glossary
       the workflow.
 
 
+   prerequisite
    dependence
    dependency
       Dependencies in the :term:`graph` show how :term:`tasks <task>` depend on
       some combination of the :term:`outputs <task output>` of other tasks.
+      These dependencies are also called the prerequisites of the downstream.
+      task.
 
       For example, in the following dependency the task ``baz`` depends on both
       ``foo`` and ``bar`` succeeding:
@@ -1461,6 +1454,8 @@ Glossary
           * :term:`task trigger`
           * :term:`conditional dependence`
           * :term:`intercycle dependence`
+
+      Tasks can also depend on :term:`external triggers <xtrigger>`.
 
 
    conditional dependence
--- a/src/glossary.rst
+++ b/src/glossary.rst
@@ -110,7 +110,7 @@ Glossary
       runtime, including tasks out to ``n`` graph edges from current
       :term:`active tasks <active task>`.
 
-      Active tasks form the ``n=0``` window.
+      Active tasks form the ``n=0`` window.
 
       .. seealso::
 
--- a/src/glossary.rst
+++ b/src/glossary.rst
@@ -76,7 +76,8 @@ Glossary
 
       Active tasks are:
 
-      - Tasks which have some, but not all of their prerequisites satisfied.
+      - Tasks which have some, but not all of their
+        :term:`prerequisites <prerequisite>` satisfied.
       - ``waiting`` tasks, that are actively waiting on:
 
         - :term:`xtriggers <xtrigger>`.
@@ -567,7 +568,7 @@ Glossary
       if the :term:`wallclock time` exceeds some offset from the cycle point.
 
       Expired is a :term:`final status` - an expired task will not run
-      even if its prerequisites get satisfied.
+      even if its :term:`prerequisites <prerequisite>` get satisfied.
 
       The associated ``:expire`` :term:`output <task output>` can be used to
       trigger other tasks. It must be marked as an :term:`optional output`,
@@ -1384,7 +1385,7 @@ Glossary
 
    prerequisite
    task prerequisite
-      A task's prerequisites are the :term:`outputs <task outputs>` of
+      A task's prerequisites are the :term:`outputs <task output>` of
       parent tasks upstream in the graph that must be completed before
       it can run. Tasks are said to "depend on" their prerequisites, hence
       the term :term:`dependency graph <graph>`.
@@ -1769,6 +1770,7 @@ Glossary
 
       If no active tasks remain and all external constraints are satisfied,
       but the n=0 window contains tasks waiting with partially satisfied
-      :term:`prerequisites`, or tasks with :term:`final status` and
-        :term:`incomplete outputs <output completion>`, then the workflow is
+      :term:`prerequisites <prerequisite>`, or tasks with :term:`final status` and
+      :term:`incomplete outputs <output completion>`, then the workflow is
       not complete and the scheduler will :term:`stall` pending manual intervention.
+
--- a/src/glossary.rst
+++ b/src/glossary.rst
@@ -70,8 +70,10 @@ Glossary
 
    active
    active task
+   active tasks
       An active task is a task which is near ready to run, in the process of
-      running, or which requires user intervention.
+      running, or which requires user intervention. These are all the tasks
+      being actively managed by the scheduler at this point in the run. 
 
       Active tasks are:
 
@@ -84,9 +86,9 @@ Glossary
 
       - ``preparing`` tasks - i.e. tasks in the process of submitting jobs
       - ``submitted`` and ``running`` tasks - i.e. those with active jobs
-      - :term: `incomplete` tasks which have completed without satisfying all of their
-        :term:`required outputs <required output>`, (e.g. a failed task
-        where success was required).
+      - tasks that finished without completing their
+        :term:`required outputs <required output>`
+        (e.g. a task failed where success was required).
 
       Active tasks are in the ``n=0`` :term:`window <n-window>` which means they
       will always be displayed in the GUI and Tui.
@@ -96,18 +98,22 @@ Glossary
 
 
    active cycle
+   active cycle point
       A cycle is active if it contains any :term:`active tasks <active task>`.
 
       Active cycles are counted towards the :term:`runahead limit`.
 
 
+   active window
+      The active window is the ``n=0`` core of the :term:`n-window`.
+      It contains the current :term:`active tasks`.
+      
+
    window
    n-window
-   active window
-   active task pool
       This is a :term:`graph`-based window or view of the workflow at runtime,
-      including tasks out to ``n`` graph edges from current :term:`active
-      tasks<active task>`. The *active window* is ``n=0``.
+      including tasks out to ``n`` graph edges from current 
+      :term:`active tasks<active task>`. The active tasks are in ``n=0``.
 
       .. seealso::
 
@@ -210,8 +216,8 @@ Glossary
    queue
    internal queue
       Internal queues (so called to distinguish them from external batch
-      queueing systems) allow you to limit how many :term:`tasks <task>` can be
-      active (submitted or running) at once, across defined groups of tasks.
+      queueing systems) limit how many :term:`jobs <job>` can be
+      active at once, across defined groups of tasks.
 
       Use queues prevent large or busy workflows from swamping their
       :term:`job platforms <job platform>` with too many jobs at once.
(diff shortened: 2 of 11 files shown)


oliver-sanders:
> task is incomplete if its outputs are incomplete, whether or not it has **finished**

As per https://github.com/cylc/cylc-admin/pull/191 (and [related discussion](https://github.com/cylc/cylc-admin/pull/190#discussion_r1536795951)), there are a couple of issues with using "finished" in this way:

* It's not technically correct as "finished" infers that the task has started which incorrectly excludes "submit-failed" and "expired".
* It conflicts with the ":finished" qualifier which is correctly defined as succeeded or failed.

Suggest continuing with "task with final status" or simply "final task" for this purpose.

---

> a stall occurs if there is nothing left to run but there are any incomplete tasks in the n=0 window

Unfortunately, there problems with this statement:

* Now, "completion" is tied entirely to "output completion", the phrase "incomplete tasks" includes active tasks. So this definition infers that a workflow can stall whilst tasks are still queued, preparing, submitted or running which is incorrect.
* Tasks with unsatisfied prerequites may actually have complete outputs (it is possible for a task not to have any required outputs at all) so "incomplete task" does not cover "task with partially satisfied prerequisites".
* Tasks waiting on xtriggers or extriggers may have incomplete outputs, but should not stall a workflow.

So "incomplete tasks" are not sufficient to define workflow stall or completion without also involving final task status.

(There are also a couple of small omissions, due to held tasks and the paused workflow status).

---

> Now, "completion" is entirely to "output completion". That means it is legit to say that a task is incomplete if its outputs are incomplete

We are ok with brining back the phrase "incomplete task" in this manner (there's no issue with the term or definition), however, the term isn't especially useful as we usually simply refer to them as "waiting tasks". I'm not sure we would find uses for this new term other than in combination with "final task status".

The much more important term is "incomplete task with a final status" for which you've used, but not defined "finished incomplete task" (which has the issues outlined above). We still prefer the original definition of "incomplete task" (including status), but failing that (ground already conceded), could we have "final incomplete task" as a defined term? We need to refer to this a lot, we also need a task event for it (https://github.com/cylc/cylc-flow/issues/4957), so could really do with a defined name or phrase.

hjoliver:
>> task is incomplete if its outputs are incomplete, whether or not it has finished

> there are a couple of issues with using "finished" in this way: ...
> ... Suggest continuing with "task with final status" or simply "final task" for this purpose.

I meant "finished" in the sense that the task has finished its automatic evolution within the workflow. Nothing else is gonna happen, short of manual intervention. That covers submit-failed and expired too.

However, I agree the semantic conflict with the finished pseudo-output is a problem.

"final task" sounds like the last task in the graph, so that won't fly.

So I guess we'll have to go with "final status".

hjoliver:
> > a stall occurs if there is nothing left to run but there are any incomplete tasks in the n=0 window

> Unfortunately, there problems with this statement:

Yeah, sorry, I didn't really mean that statement in the description, as opposed to the docs, to exhaust all the possibilities. It was implicit (in my mind at least :-) that nothing is currently running either.  

However good point on tasks with no required outputs.

hjoliver:
> We are ok with brining back the phrase "incomplete task" in this manner (there's no issue with the term or definition), however, the term isn't especially useful as we usually simply refer to them as "waiting tasks".

No, it's not just waiting tasks - also submitted and running ones that haven't completed their outputs yet.

hjoliver:
>  We still prefer the original definition of "incomplete task" (including status), but failing that (ground already conceded), could we have "final incomplete task" as a defined term?

We've covered that ground, as you say. I do think that "incomplete task" as we originally used it (final status with incomplete outputs) was nice in isolation, but in the wider context it didn't make any sense that a task wasn't incomplete *until* it achieved final status - so what on earth was its "completion status" prior to becoming incomplete? (I'm not even going to say "in my opinion" on that one, because it is a self-evident truth).

> could we have "final incomplete task" as a defined term? We need to refer to this a lot, we also need a task event for it (https://github.com/cylc/cylc-flow/issues/4957), so could really do with a defined name or phrase.

Yes, it would be good to have a concise term for the concept "incomplete task with final status". It's a pity that we can't use "finished" as I meant it above, because of the pseudo output, as "finished incomplete" is pretty intuitive and concise.

I don't like "final incomplete task" much because it sounds like "the last incomplete task".

Perhaps we could invent a new term, divorced from all the baggage that comes with the words "final" and "complete"?

An idea: **zombie task** 
   - it achieved final status (end of life, short of manual intervention) but can't leave the task pool (the world of the living) because it is incomplete
   - (too ridiculous?)
 
(But I'll fix up the text in this PR to use the full definition for now).

hjoliver:
@oliver-sanders - I've amended the text to fix up "finished" and "final status" as discussed above.  The build fail is a link-check.

src/glossary.rst
```diff
@@ -1694,11 +1740,11 @@ Glossary
    runahead limit
    runahead
       In a :term:`cycling workflow`, the runahead limit determines how far
-      ahead of the oldest :ref:`active cycle point` the workflow is permitted
+      ahead of the oldest :term:`active cycle point` the workflow is permitted
```
oliver-sanders, line 1743: Personal preference (cycles are a clearer concept without the "point" bit):

```suggestion
      ahead of the oldest :term:`active cycle` the workflow is permitted
```

src/glossary.rst
```diff
@@ -1708,12 +1754,12 @@ Glossary
 
 
    workflow completion
-      A workflow is deemed complete if there are no more tasks to run,
-      according to the graph, and there are no :term:`incomplete task
-      <incomplete task>` present.
+      A workflow is complete if there are no more tasks to run according
+      to the graph, and there are no :term:`incomplete <output completion>`
+      tasks present in the :term:`n=0 window <n-window>`.
```
oliver-sanders, line 1759: Incomplete outputs are not sufficient to diagnose workflow completion.

This definition needs to include final task status.

src/user-guide/running-workflows/reflow.rst
```diff
@@ -128,38 +128,26 @@ Special Case: Triggering ``n=0`` Tasks
 
    - Triggering a task with a submitted or running job has no effect
      (it is already triggered).
-   - Triggering other :term:`active tasks <active task>` e.g. (a waiting
-     task which is held) queues it to run in the same flow.
-   - Triggering an :term:`incomplete task` queues it to re-run in the same flow.
+   - Triggering other :term:`active tasks <active task>`, including
+     :term:`incomplete <output completion>` tasks with a :term:`final status`,
+     queues them to run in the same flow.
```
oliver-sanders, line 133: I don't think "final status" actually comes into this one:

```suggestion
   - Triggering other :term:`active tasks <active task>`, including
     tasks with :term:`incomplete outputs <output completion>`,
     queues them to run in the same flow.
```

src/user-guide/running-workflows/reflow.rst
```diff
@@ -128,38 +128,26 @@ Special Case: Triggering ``n=0`` Tasks
...
-expectation of retriggering to complete :term:`required outputs<required
-output>` and continue their flow.
-
-If another flow encounters an incomplete task (i.e. if another instance of the
-same task collides with it in the ``n=0`` :term:`active window`) the task will
-run again and carry both flow numbers forward if it successfully completes its
-required outputs.
+If original task instance has a :term:`final status` but is
+:term:`incomplete <output completion>` the merged task will be reset to
+run again without manual intervention.
```
oliver-sanders, line 150: I don't think "final status" actually comes into this one:

```suggestion
If the original task instance has :term:`incomplete outputs`,
the merged task will be reset to
run again without manual intervention.
```

src/user-guide/running-workflows/workflow-completion.rst
```diff
@@ -4,25 +4,29 @@ Workflow Completion
 ===================
 
 If there is nothing more to run (according to the graph) and there are no
-:term:`incomplete tasks <incomplete task>` present, the scheduler will report
-workflow completion and shut down when current active tasks finish. 
+:term:`incomplete <output completion>` tasks present in :term:`n=0 <n-window>`
+the workflow is complete and the scheduler will shut down automatically.
```
oliver-sanders, line 8: This definition isn't correct (see other comment in the glossary page).

src/user-guide/running-workflows/workflow-completion.rst
```diff
@@ -4,25 +4,29 @@ Workflow Completion
...
 Scheduler Stall
 ===============
 
-If there is nothing more to run (according to the graph) but there are
-:term:`incomplete tasks <incomplete task>` present, the scheduler will
-:term:`stall` and stay alive for 1 hour (by default) awaiting user intervention
+If there is nothing more to run, but there are
+:term:`incomplete <output completion>` tasks present in the
+:term:`n=0 window <n-window>` the workflow did not run to
+completion, so the scheduler will :term:`stall` and stay
```
oliver-sanders, line 19: This definition isn't correct (see other comment in the glossary page).

oliver-sanders (commented):
> [from the OP]
>
> E.g. it means we can say a stall occurs if there is nothing left to run but *there are any incomplete tasks in the n=0 window"

As noted, unfortunately, this is not true. Suggest not re-adding this term (not especially useful).

It might make it easier to define "final incomplete" though as this is a top-level concept that users need to be aware of.

src/glossary.rst
```diff
@@ -95,18 +97,22 @@ Glossary
 
 
    active cycle
+   active cycle point
```
oliver-sanders, line 100: Pairs with previous suggestion:

```suggestion
```

hjoliver:
(Sorry, this got lost in the workflow-state PR mess - will come back to it ASAP ... I think I had already agreed with several of your points above but had come back to change the text yet).

src/glossary.rst
```diff
@@ -70,8 +70,10 @@ Glossary
 
    active
    active task
+   active tasks
```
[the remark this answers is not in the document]
hjoliver: (Yep, not sure why I did that).

src/7-to-8/major-changes/compatibility-mode.rst
```diff
@@ -34,7 +34,8 @@ The ``suite.rc`` filename triggers a backward compatibility mode in which:
     branching)
 
 - only ``succeeded`` task outputs are :ref:`*required* <User Guide Required Outputs>`,
-  meaning the scheduler will retain tasks that do not succeed as incomplete
+  so the scheduler will retain all non-succeeded :term:`finished tasks <finished task>`
+  in the :term:`n=0 window <n-window>`.
```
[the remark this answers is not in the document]
hjoliver: (compromise explanation used)

src/glossary.rst
```diff
@@ -95,18 +97,22 @@ Glossary
...
    active cycle
+   active cycle point
       A cycle is active if it contains any :term:`active tasks <active task>`.
 
       Active cycles are counted towards the :term:`runahead limit`.
 
 
+   active window
+      The active window is the ``n=0`` core of the :term:`n-window`.
+      It contains the current :term:`active tasks`.
```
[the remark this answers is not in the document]
hjoliver: Good idea.

src/glossary.rst
```diff
@@ -1555,16 +1601,16 @@ Glossary
...
       User intervention is required to fix a stall, e.g. by retriggering
-      incomplete tasks after fixing the problems that caused them to fail.
+      tasks after fixing the problems that caused them to fail.
 
 
    suicide trigger
-      Suicide triggers remove :term:`tasks <task>` from the :term:`scheduler's
-      <scheduler>` active (:term:`n=0 <n-window>`) window at runtime.
+      Suicide triggers remove :term:`tasks <task>` from the 
+      :term:`n=0 window <n-window>` at runtime.
```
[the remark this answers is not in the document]
hjoliver: Sorry, not sure I follow you?

src/glossary.rst
hjoliver: I prefer "cycle point" for clarity - it distinguishes between the sequence as a whole (e.g., a "yearly cycle") and the point value ("2034"). 

Also, I think we still have a few user-facing (albeit possibly deprecated) options and arguments called "point" rather than "cycle".

I'll concede that sometimes, as in this example, "point" can be probably be omitted because the context makes it clear.

src/glossary.rst
hjoliver: If you look at the full sentence, it is very close to correct without listing all the cases. By "no more tasks to run" I include tasks that are ready but held back by (e.g.) xtriggers; and I assumed that none currently running was obvious. Partially satisfied tasks with no required outputs are perhaps the sticking point.

I'll avoid duplication by referring to the glossary entry for "stalled" instead, and make sure that's correct.

src/user-guide/running-workflows/reflow.rst
hjoliver: I only said "including" - because incomplete final-status tasks are the main target for retriggering by far. But I'll accept your change...

src/user-guide/running-workflows/reflow.rst
hjoliver: It does in this case. A task can only be "reset to run again" if it has already achieved a final status.

src/user-guide/running-workflows/workflow-completion.rst
hjoliver: reworded

src/user-guide/running-workflows/workflow-completion.rst
hjoliver: reworded

hjoliver:
All review comments addressed, but I've run out of time to proof-read my changes or do a build check.

src/glossary.rst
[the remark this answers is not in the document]
oliver-sanders: Since https://github.com/cylc/cylc-flow/pull/6067, suicide triggers don't *just* remove tasks from the task pool (aka the n=0 window), because the "removal" is indirectly recorded in the database (i.e. the task is considered "complete", but with a non-final status :thinking:).

I think a task that is suicide-triggered should be auto-spawned anyway so the n=0 bit is redundant. The limitation of suicide triggers is that they cant remove a task *after* it has become active.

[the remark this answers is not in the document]
hjoliver: Oh, right, gotcha.  Well, perhaps it's kind of a matter of perspective. Only tasks in n=0 can cause a stall, so a task does not need to be suicided at all unless (a) it is gets spawned into n=0, and (b) would cause a stall if not removed.  So, they remove tasks from n=0, right?

> I think a task that is suicide-triggered should be auto-spawned anyway so the n=0 bit is redundant.

Do you mean auto-spawned by the suicide dependency? (and then, immediately removed by the suicide trigger).  If so, that's just an implementation trick to allow continued support for suicide triggers in places where they aren't actually needed at all in Cylc 8.

hjoliver:
Hopefully good to go now @oliver-sanders

hjoliver:
The final small commit tweaks the docs for workflow-state changes https://github.com/cylc/cylc-flow/pull/5809

src/workflows/xtrigger/workflow_state/downstream/flow.cylc
```diff
@@ -3,8 +3,7 @@
 [scheduling]
     initial cycle point = 2010
     [[xtriggers]]
-         upstream = workflow_state(workflow=up, task=foo, point=%(point)s, \
-            message='data ready'):PT10S
+         upstream = workflow_state(up//%(point)s/foo:x):PT10S
```
oliver-sanders, line 6: It would be helpful for https://github.com/cylc/cylc-flow/issues/5588 if we quoted xtrigger arguments in the documentation:

```suggestion
         upstream = workflow_state("up//%(point)s/foo:x"):PT10S
```

src/workflows/xtrigger/sequential/flow.cylc
```diff
@@ -4,8 +4,7 @@
     initial cycle point = 2010
     sequential xtriggers = True
     [[xtriggers]]
-         upstream = workflow_state(workflow=up, task=foo, point=%(point)s, \
-            message='data ready'):PT10S
+         upstream = workflow_state(up//%(point)s/foo:x):PT10S
```
oliver-sanders, line 7: It would be helpful for https://github.com/cylc/cylc-flow/issues/5588 if we quoted xtrigger arguments in the documentation:

```suggestion
         upstream = workflow_state("up//%(point)s/foo:x"):PT10S
```

src/user-guide/writing-workflows/scheduling.rst
```diff
@@ -1528,29 +1536,31 @@ in :ref:`Section External Triggers`.
...
+intervention, e.g. to be retriggered after a bug fix.
 
 .. note::
-   Incomplete tasks stall the workflow if there is nothing else to do (see
-   :ref:`workflow completion`). They also count toward the :term:`runahead
-   limit`, because they may run again once dealt with.
+   Tasks that achieve a final status without completing their outputs will
+   raise a warning and stall the workflow when there is nothing else for
+   the scheduler to run (see :ref:`workflow completion`). They also count
+   toward the :term:`runahead limit`, because they may run again once dealt with.
```
oliver-sanders, line 1563: > because they may run again once dealt with.

???

```suggestion
   toward the :term:`runahead limit`.
```

oliver-sanders (commented):
(just minor stuff)

src/glossary.rst
```diff
@@ -96,18 +98,19 @@ Glossary
...
 
 
    window
    n-window
    active window
-   active task pool
-      This is a :term:`graph`-based window or view of the workflow at runtime,
-      including tasks out to ``n`` graph edges from current :term:`active
-      tasks<active task>`. The *active window* is ``n=0``.
+      The UI provides a :term:`graph`-based window or view of the workflow at
```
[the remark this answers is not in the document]
hjoliver: I kinda meant "the UIs" in that it's the datastore, potentially for any UI (albeit not Tui yet), but I'll change it for now.

src/glossary.rst
```diff
@@ -1346,20 +1367,39 @@ Glossary
...
 
+   prerequisite
+   task prerequisite
+      A task's prerequisites are the :term:`outputs <task output>` of
+      parent tasks upstream in the graph that must be completed before
+      it can run. Tasks are said to "depend on" their prerequisites, hence
+      the term :term:`dependency graph <graph>`.
+
+      Tasks can depend on :term:`external triggers <xtrigger>` as well as
+      the outputs of other, upstream, tasks.
```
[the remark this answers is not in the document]
hjoliver: Good spot, I've combined the two.

src/user-guide/writing-workflows/scheduling.rst
hjoliver: The point is, even though they have final status, they are not done with as far as the workflow is concerned and may well get run again by manual intervention - which explains why they should count toward the runahead limit.

However, I agree that's not a very elegant way of putting it!

src/user-guide/writing-workflows/external-triggers.rst
```diff
@@ -402,6 +395,7 @@ success. The function signature is:
 .. automodule:: cylc.flow.xtriggers.xrandom
    :members: xrandom, validate
    :member-order: bysource
+   :noindex:
```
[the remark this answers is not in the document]
hjoliver: which has indeed returned:
```
/home/oliverh/cylc/cylc-doc/venv/lib/python3.8/site-packages/cylc/flow/xtriggers/xrandom.py:docstring
     of cylc.flow.xtriggers.xrandom:1: 
WARNING: duplicate object description of cylc.flow.xtriggers.xrandom, 
    other instance in plugins/xtriggers/built-in/cylc.flow.xtriggers.xrandom, use :noindex: for one of them
```

hjoliver:
The one build problem is fixed on https://github.com/cylc/cylc-flow/pull/5809

```
 <unknown>:1556: WARNING: term not in glossary: 'incomplete'
```

src/glossary.rst
```diff
@@ -1381,10 +1410,37 @@ Glossary
...
+      the workflow.
+
+
+   prerequisite
    dependence
    dependency
       Dependencies in the :term:`graph` show how :term:`tasks <task>` depend on
       some combination of the :term:`outputs <task output>` of other tasks.
+      These dependencies are also called the prerequisites of the downstream.
+      task.
```
MetRonnie, line 1443:
```suggestion
      These dependencies are also called the prerequisites of the downstream
      task.
```

Or possibly get rid of this, as I find the wording a little confusing, and I think the definition of prerequisite is already clear.

```suggestion
```

src/glossary.rst
```diff
@@ -106,11 +106,11 @@ Glossary
    window
    n-window
    active window
-      The UI provides a :term:`graph`-based window or view of the workflow at
+      The GUI provides a :term:`graph`-based window or view of the workflow at
       runtime, including tasks out to ``n`` graph edges from current
       :term:`active tasks <active task>`.
 
-      Active tasks form the ``n=0`` core of the :term:`n-window`.
+      Active tasks form the ``n=0``` window.
```
oliver-sanders, line 113:
```suggestion
      Active tasks form the ``n=0`` window.
```

oliver-sanders (approved):

src/user-guide/writing-workflows/external-triggers.rst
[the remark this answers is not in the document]
MetRonnie: I don't get that when building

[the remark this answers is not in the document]
oliver-sanders: Different Sphinx versions?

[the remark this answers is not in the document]
hjoliver: Probably. So let's just ignore the warning in my env for now. 
(Note if I add `:noindex:` my warning goes away, but then @MetRonnie gets the one above that I don't get!)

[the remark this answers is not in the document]
MetRonnie: I can't see any duplicate autodoc of `cylc.flow.xtriggers.xrandom`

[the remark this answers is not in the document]
oliver-sanders: A duplicate index will cause a Sphinx error (which is what the no-index fiag is for).

oliver-sanders:
Just the one (expected) build error: https://github.com/cylc/cylc-doc/actions/runs/9497881655/job/26175443458?pr=727#step:10:320

MetRonnie (approved):

PR #733, opened 2024-05-17
changes: add group by cycle point entry

oliver-sanders:
(note tests currently failing on master due to a couple of referencing issues, Hillary has a PR open to resolve this)

**Requirements check-list**


src/reference/changes.rst
```diff
@@ -120,6 +123,15 @@ toolbar allowing you to see tasks further back in the workflow's history.
    Using high n-window values with complex workflows may have performance
    impacts.
 
+
+Group by cycle point in the graph view
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The graph view now has an option to group tasks by cycle point.
```
MetRonnie, line 130:
```suggestion
.. versionadded: cylc-uiserver 1.5.0

The graph view now has an option to group tasks by cycle point.
```

MetRonnie (approved):

src/reference/changes.rst
oliver-sanders: We only include versionadded entries if the feature was introduced in a maintenance release (which shouldn't normally happen, but we are a bit cheeky about sneaking in cylc-ui features as they are less likely to be dissruptive).

PR #739, opened 2024-06-14
interventions: document holding all tasks, triggering and releasing

oliver-sanders:
Document a common use case:
* Hold all tasks
* Trigger some tasks
* Release all tasks

Currently the solution is to use the "hold after cycle point" feature.

Note, when https://github.com/cylc/cylc-flow/issues/5827 is done, we will be able to simplify this to:

```
$ cylc hold 'workflow//*'
$ cylc trigger ...
$ cylc release 'workflow//*'
```

**Requirements check-list**


src/user-guide/interventions/index.rst
```diff
@@ -439,3 +439,52 @@ Terminate a Chain of Automatic Retries
...
+         $ # release the "hold point" to allow the workflow to continue
+         $ cylc release --all <workflow>
+
+.. note::
+
+   The difference between the workflow "paused" state and the task "held" state:
+
+   Workflow Pause
+      When a workflow is :term:`paused <pause>`, no new jobs will be submitted.
+      This gives you an oportunity to make changes to the workflow, however, if
```
markgrahamdawson, line 485:
```suggestion
      This gives you an opportunity to make changes to the workflow, however, if
```

src/user-guide/interventions/index.rst
```diff
@@ -439,3 +439,52 @@ Terminate a Chain of Automatic Retries
...
+   * Trigger the task(s) you want to run.
+   * When you're done triggering, release the held tasks to allow the workflow
+     to continue.
+
+.. tab-set::
+
+   .. tab-item:: GUI
+      :sync: gui
+
+      .. image:: set-and-release-hold-point.gif
```
MetRonnie, line 462: This is quite a lengthy GIF which could be halfway through playback when the user starts trying to follow along with it. Is it possible to embed a video instead?

oliver-sanders: Dunno.

oliver-sanders: Run the GIF through once, then stop and provide a button to play from the start is another option.

oliver-sanders:
(merging with two approvals)

PR #748, opened 2024-06-25
Allow symlinking for log/job

wxtim:
Accompanies https://github.com/cylc/cylc-flow/pull/6168

**Requirements check-list**

### DIFF
--- a/src/reference/config/writing-platform-configs.rst
+++ b/src/reference/config/writing-platform-configs.rst
@@ -287,7 +287,7 @@ locations. The entire :term:`run directory` can be symlinked, as well as
 certain sub-directories.
 
 * ``run`` - the :term:`run directory` itself
-* ``log`` (see :term:`workflow log directory`)
+* ``log``
   * ``log/job`` - contains job scripts and outputs.
 * ``share`` (see :term:`share directory`)
 * ``share/cycle`` (typically used by `Rose`_ tasks)
--- a/src/reference/config/writing-platform-configs.rst
+++ b/src/reference/config/writing-platform-configs.rst
@@ -288,10 +288,7 @@ certain sub-directories.
 
 * ``run`` - the :term:`run directory` itself
 * ``log`` (see :term:`workflow log directory`)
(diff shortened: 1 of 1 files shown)


src/reference/config/writing-platform-configs.rst
```diff
@@ -287,7 +287,11 @@ locations. The entire :term:`run directory` can be symlinked, as well as
 certain sub-directories.
 
 * ``run`` - the :term:`run directory` itself
-* ``log``
+* ``log`` (see :term:`workflow log directory`)
+* ``log/job`` The ``log/job`` directory contains more files than ``log``
+  and the risks (to the workflow) from losing them are smaller: Tasks
+  without log files might need re-running: Loss of workflow logs might
+  require re-running part or all of a workflow.
```
oliver-sanders, line 294: > ``log/job`` directory contains more files than ``log``

the job/log dir is contained within log

> Tasks  without log files might need re-running

Data is generally created in work or share not log so loss of logs is not usually a concern.



---

I think we can strip the extra info, if we are going to add this sort of content, we should do it for all the dirs.


```suggestion
  * ``log/job`` - contains job scripts and outputs.
```

wxtim:
Can I poke @oliver-sanders, @markgrahamdawson   - I think it's a single reviewer thing.

src/reference/config/writing-platform-configs.rst
```diff
@@ -287,7 +287,8 @@ locations. The entire :term:`run directory` can be symlinked, as well as
 certain sub-directories.
 
 * ``run`` - the :term:`run directory` itself
-* ``log``
+* ``log`` (see :term:`workflow log directory`)
```
oliver-sanders, line 290: The glossary appears to define "workflow log directory" as `log/scheduler`: https://cylc.github.io/cylc-doc/stable/html/glossary.html#term-workflow-log-directory

```suggestion
* ``log``
```

oliver-sanders (approved):

PR #757, opened 2024-08-15
whilst -> while

hjoliver:
As flagged in Element chat, "whilst" is entirely equivalent to "while" in all uses here, but "whilst" sticks out as old-fashioned to many readers. We mostly use "while" in the docs, but a few "whilst"s have crept in.

**Requirements check-list**
- [ ] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.


MetRonnie (approved):
Spellcheck failure unrelated

PR #772, opened 2024-11-07
document graph branching with multiple optional outputs

oliver-sanders:
A topic which periodically comes up but is not presently documented.

See also: https://github.com/cylc/cylc-flow/issues/5729

**Requirements check-list**

### DIFF
--- a/src/user-guide/writing-workflows/scheduling.rst
+++ b/src/user-guide/writing-workflows/scheduling.rst
@@ -1909,6 +1909,108 @@ A more realistic example might have several tasks on each branch. The
 ``bar``, but configured differently to avoid the failure.
 
 
+Dependencies With Multiple Optional Outputs
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+We might have a task that depends on multiple optional outputs.
+
+For example, this workflow is like the "recovery task" example above, but in
+this case we only want to run the recover task if both of the upstream tasks
+fail.
+
+.. digraph:: Example
+   :align: center
+
+   rankdir = "LR"
+
+   one -> run_if_both_fail [label=":fail?", color="red", "fontcolor"="red"]
+   two -> run_if_both_fail [label=":fail?", color="red", "fontcolor"="red"]
+
+   one -> always_run [label=":finish", color="orange", "fontcolor"="orange"]
+   two -> always_run [label=":finish", color="orange", "fontcolor"="orange"]
+
+   subgraph cluster_1 {
+     label = "one:failed AND two:failed"
+     color = "red"
+     fontcolor = "red"
+     style = "dashed"
+
+     run_if_both_fail
+   }
+
+   subgraph cluster_2 {
+     label = "both tasks finished"
+     color = "orange"
+     fontcolor = "orange"
+     style = "dashed"
+
+     always_run
+   }
+
+We might try to write the graph like so:
+
+.. code-block:: cylc-graph
+
+   # run irrespective of whether the tasks succeed or fail
+   one:finish & two:finish => always_run
+
+   # run if both tasks fail  <-- ERROR
+   one:fail? & two:fail? => run_if_both_fail
+
+However, there is a problem with this.
+
+1. If both tasks fail, then ``run_if_both_fail`` will run.
+2. If both tasks succeed, then ``run_if_both_fail`` will not run.
+3. If one task succeeds and the other fails, then the task ``run_if_both_fail``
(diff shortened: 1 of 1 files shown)


src/user-guide/writing-workflows/scheduling.rst
```diff
@@ -1909,6 +1909,108 @@ A more realistic example might have several tasks on each branch. The
...
+
+However, there is a problem with this.
+
+1. If both tasks fail, then ``run_if_both_fail`` will run.
+2. If both tasks succeed, then ``run_if_both_fail`` will not run.
+3. If one task succeeds and the other fails, then the task ``run_if_both_fail``
+   will be left with one satisfied and one unsatisfied dependency. This will
+   cause the workflow to :term:`stall`.
+
+To prevent the workflow from stalling in the third case, it is necessay to use
```
hjoliver, line 1968:
```suggestion
To prevent the workflow from stalling in the third case, it is necessary to use
```

src/user-guide/writing-workflows/scheduling.rst
```diff
@@ -1909,6 +1909,108 @@ A more realistic example might have several tasks on each branch. The
...
+
+   # run irrespective of whether the tasks succeed or fail
+   one:finish & two:finish => always_run
+
+   # run if both tasks fail
+   one:fail? & two:fail? => run_if_both_fail
+   one:succeeded? | two:succeeded? => !run_if_both_fail
+
+Here's an example workflow showing how to trigger tasks with each possible
+combination of success/failre for the two tasks:
```
hjoliver, line 1982:
```suggestion
combination of success/failure for the two tasks:
```

hjoliver (changes requested):
Two minor typos.

Personal preference that I'm not super-invested in, but see what you think:

```
one:fail? & two:fail? => run_if_both_fail
```
I suppose this reduces verbosity by conveying additional information in the task name,  but there's a risk that new users might think the name is magic (i.e., functional in some way) or is something other than a simple task name (it's a weird task name after all).  IMO this would be better:

```
one:fail? & two:fail? => three  # run three if both one and two fail
```

wxtim (approved):
Good. Please fix the spelling errors Hilary spotted.

oliver-sanders:
> Two minor typos.
>
> Personal preference that I'm not super-invested in, but see what you think:

Usually these review comments go the other way, preferring concrete to abstract names.

Especially given the context, I don't think people will think these names are "magic", e.g. there's "recover" task in the previous example and an "ugly" task in the next one.

hjoliver:
Well, "recover" conveys what the task does (/what its purpose is), which kinda makes it a normal task name. And abstract task names are fine when their purpose doesn't matter for the sake of example. 

"run_if_both_fail" is different, and redundant actually, because it spells out its own triggering condition which is already evident in the graph. 

Plus the other task names in the example are abstract and sensibly related to "three".

oliver-sanders:
(will leave you two to battle that out)

hjoliver:
(but @wxtim hasn't given an argument one way or the other on this .. it seems you and I have to decide!)

wxtim:
In my opinion none of us can decide who is right on this because we are too close to the problem and we should either accept it or go and test it on a user.

hjoliver:
I rather like that description (running mental models of users in our heads)!  ... but asking a real user, or even a bunch of them, would not really be sufficient. There are many different kinds of user, and we can't run surveys for every little detail of this nature. So unfortunately we have to rely on our mental models for most of this stuff.

The task name `run_if_both_fail` is potentially confusing because it restates its own triggering condition. That is a strange way to name a task. Some readers may find that puzzling and may even wonder if it has some kind of functional meaning. And there's no real value in it because the information in the name is redundant - it's in the graph trigger already! 

The only counter-argument has been (paraphrasing) "I don't think it's confusing". 

A generic task name, with a description of the triggering in the text if we like, has zero confusion potential. We should always go with zero potential for confusion, right? 

Generic task names are good *when it doesn't matter what the task does in the context of the example*. They require no mental effort from the user. (Plus the other tasks in the same example are already generic!)

hjoliver:
@wxtim - I see @oliver-sanders has removed his assignment from the PR, despite being the author!

I presume that reflects his dislike for terminology discussions and similar, and doesn't care one way or the other.
(which is fair enough).

I do care one way or the other, and have given (IMO) good reasons for my stance.

 Do you have a good counter-argument?  If not, I'll push a change up...

If not, I guess I'll just push a change the branch...

hjoliver (approved):

PR #778, opened 2024-11-21
Update docs for trigger-when-paused.

hjoliver:
Companion for https://github.com/cylc/cylc-flow/pull/6192

@oliver-sanders - I still need to update the intervention gif animiation. Is it possible for you to advise on window size etc. so that I end up with a similar look to your originals, or do I just have to experiment with that?

TODO:
- [ ] update the intervention gif animation

**Requirements check-list**

### DIFF
--- a/src/glossary.rst
+++ b/src/glossary.rst
@@ -1151,7 +1151,9 @@ Glossary
 
    pause
       When a :term:`workflow` is "paused" the :term:`scheduler` is still
-      running but it will not submit any new jobs.
+      running but it will not automatically submit new jobs.
+
+      You can still :term:`trigger` jobs manually in a paused workflow.
 
       This can be useful if you want to make a change to a running workflow.
 
@@ -1486,6 +1488,9 @@ Glossary
       Triggers can be based on :term:`standard <standard output>` or
       :term:`custom <custom output>` task outputs. In the latter case they
       are known as :term:`message triggers <message trigger>`.
+
+      The term "trigger" also refers the action of manually triggering
+      a task to run, via the ``cylc trigger`` command or the GUI. 
 
 
    message trigger
--- a/src/user-guide/interventions/index.rst
+++ b/src/user-guide/interventions/index.rst
@@ -441,18 +441,17 @@ but specify the "failed" output rather than
 "succeeded".
 
 
-Hold The Workflow And Trigger Tasks One By One
-----------------------------------------------
+Pause The Workflow And Trigger Tasks One By One
+-----------------------------------------------
 
 :Example:
-   I want to hold back the workflow while I manually run one or more tasks
+   I want to pause the workflow while I manually run one or more tasks
    to fix a problem or test a task.
 
 :Solution:
-   * Hold all tasks after a specified cycle.
+   * Pause the workflow.
    * Trigger the task(s) you want to run.
-   * When you're done triggering, release the held tasks to allow the workflow
-     to continue.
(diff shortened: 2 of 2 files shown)


oliver-sanders:
> @oliver-sanders - I still need to update the intervention gif animiation. Is it possible for you to advise on window size etc. so that I end up with a similar look to your originals, or do I just have to experiment with that?

Some details in this comment:

https://github.com/cylc/cylc-doc/blob/8afb75de3e99f0a55876ee976189c9ee1e55bdae/src/user-guide/interventions/index.rst?plain=1#L48-L55

Something I didn't mention there, but should have, is the terminal theme I used (Gnome - Tango Dark) for Tui recordings.

src/glossary.rst
```diff
@@ -1151,7 +1151,11 @@ Glossary
 
    pause
       When a :term:`workflow` is "paused" the :term:`scheduler` is still
-      running but it will not submit any new jobs.
+      running but it will not automatically submit new jobs.
+
+      You can still manually :term:`trigger` tasks in a paused workflow. By
+      default they will run once the workflow is resumed, or immediately
+      with the ``--now`` trigger option.
```
oliver-sanders, line 1158: Needs updating now the default has switched.

src/glossary.rst
```diff
@@ -1247,7 +1251,11 @@ Glossary
          :term:`Tasks <task>` can be :term:`held <hold>` with ``cylc hold`` and
          :term:`released <release>` with ``cylc release``.
 
-         When a workflow is resumed, any held tasks remain held.
+         When a paused workflow is resumed, any held tasks remain held.
+
+         If a running task is held its running job is not affected, but it will
+         not submit any more jobs (such as automatic :term:`retries <retry>`)
+         until released. The children of held tasks are not automatically held.
```
oliver-sanders, line 1258: I get what you mean here (children = graph children), but I don't think users will.

We can omit this (there is no reason to think that this might happen):

```suggestion
         until released. 
```

src/user-guide/interventions/index.rst
```diff
@@ -441,18 +441,17 @@ but specify the "failed" output rather than
...
+   I want to pause the workflow while I manually run one or more tasks
    to fix a problem or test a task.
 
 :Solution:
-   * Hold all tasks after a specified cycle.
-   * Trigger the task(s) you want to run.
-   * When you're done triggering, release the held tasks to allow the workflow
-     to continue.
+   * Pause the workflow.
+   * Trigger the task(s) you want to run, with the ``--now`` option.
```
oliver-sanders, line 453: Switcharoo

oliver-sanders:
Could also do with a quick entry in the [changes section](https://cylc.github.io/cylc-doc/stable/html/reference/changes.html), we'll post this on Discourse, etc to advertise the change in behavior.

src/glossary.rst
hjoliver: Agreed on graph children.

Less agreed on "there's no reason to think..." - I've had it several times from users who tried to "hold everything" by holding "all the tasks" - which doesn't hold everything because the children of any active tasks will carry on.

However, I'll take your suggestion anyway.

hjoliver:
Text review suggestions addressed.

TODO:
- example animation
- add to the  [changes section](https://cylc.github.io/cylc-doc/stable/html/reference/changes.html)

wxtim:
https://github.com/hjoliver/cylc-doc/pull/3

MetRonnie:
Superseded by #790

PR #789, opened 2025-01-07
added a remove-task intervention to the user guide

### LINKED ISSUE #785: interventions: removing tasks
Document an [intervention](https://cylc.github.io/cylc-doc/stable/html/user-guide/interventions/index.html) for removing tasks.

This use case is written up in cylc-admin: https://github.com/cylc/cylc-admin/blob/master/docs/proposal-interventions.md#1-i-triggered-tasks-i-didnt-mean-to-they-may-have-run-on-i-want-to-undo-this

wxtim:
closes #785 

**Requirements check-list**


src/user-guide/interventions/index.rst
```diff
@@ -523,3 +523,38 @@ I want to Skip a cycle of tasks and allow the workflow to continue
...
+
+   .. tab-item:: GUI
+      :sync: gui
+
+      .. image:: remove.gif
+         :width: 75%
+
+      .. warning::
+
+         Internal implemenation details mean that the task icon will not be
```
wxtim, line 548: Anyone think of a better way to describe this?

hjoliver: I definitely wouldn't say it as you've written it - it makes it sound as if the GUI has unfixed bug or just can't represent the reality of the situation. In addition to noting the greying out, I think we should say something like this to indicate that the potential visibility of "removed" tasks is deliberate and correct:

```
Removing a task might not make it disappear from the GUI, because the
GUI displays all tasks in a graph-based window around current active tasks.
```

src/user-guide/interventions/index.rst
```diff
@@ -523,3 +523,38 @@ I want to Skip a cycle of tasks and allow the workflow to continue
...
+         removed.
+
+         .. the implementation detail in question is that remove removes
+         .. flows from tasks.
```
hjoliver, line 552:
```suggestion
```

hjoliver, line 552:
```suggestion
         The removed task will be greyed out but it might not
         disappear from view because the GUI displays the graph
         around current active tasks.

         .. the implementation detail in question is that remove removes
         .. flows from tasks.
```

src/user-guide/interventions/index.rst
hjoliver: (Note Ronnie's grey-out (flow-nums) branch is now merged)

MetRonnie: Yes, I suggest changing to a note, and I endorse ~Hilary for president~ Hilary's suggestion below. The GIF ought to be re-recorded

src/user-guide/interventions/index.rst
```diff
@@ -523,3 +523,36 @@ I want to Skip a cycle of tasks and allow the workflow to continue
...
+
+.. tab-set::
+
+   .. tab-item:: GUI
+      :sync: gui
+
+      .. image:: remove.gif
+         :width: 75%
+
+      .. warning::
```
MetRonnie, line 546:
```suggestion
      .. note::
```

src/user-guide/interventions/remove.gif
MetRonnie, line 1: The workflow example you have given leads me to think the user would want to remove `dont_run_me_either` as well

MetRonnie: I think if the example would require removing 2 tasks it should show removing 2 tasks?

wxtim: Go on then.

oliver-sanders: I'm not getting the task names either.

oliver-sanders: If removing two tasks, it might be worth showing it remove a running task as well as a succeeded one.

MetRonnie (approved):

MetRonnie (approved):

oliver-sanders (approved):

PR #790, opened 2025-01-08
Trigger when paused

wxtim:
Copy of https://github.com/cylc/cylc-doc/pull/778 with spelling fix and updated animation, but not reference/changes, all of which I'm going to do as a single PR to avoid need to deconflict

**Requirements check-list**


src/dictionaries/words
```diff
@@ -209,6 +209,7 @@ toolbar
 tuple
 uiserver
 unparsed
+unpause
 unpaused
```
MetRonnie, line 213:
```suggestion
```
IMO we should be consistently using the word "resume", not "unpause"

src/user-guide/interventions/trigger-while-paused.gif
MetRonnie, line 1: Not a big deal, but I think this would be more "concise" if the workflow was running when the GIF starts

oliver-sanders: I have no idea what's causing that "yellowing" strange (custom CSS / browser settings?).

I think this example could be a tad simpler, the trigger operation is kinda lost in play/pause/graph-pan operations. Suggest:

* Use the tree view and a workflow simple enough to fit on one page (no panning/scrolling).
* Start with the workflow running (as suggested above).
* Pause the workflow and observe that tasks aren't running.
* Trigger one task and watch it run.

src/dictionaries/words
wxtim: Objection - Hilary has used this word in a way which clarifies rather than obfuscates.

oliver-sanders: IMO both are fine.

src/user-guide/interventions/trigger-while-paused.gif
wxtim: Yellowing is my desktop theme.

It is playing to start with.

wxtim: Updated.

oliver-sanders (approved):

MetRonnie (approved):

PR #799, opened 2025-02-18
fix build error

oliver-sanders:
* Sphinx tries to cross-link typing information.
* If it can't find a type to link to, it will fail in nit-pick mode.
* If we don't want it to fail, we must list the references in an ignore section.

Example failure: https://github.com/cylc/cylc-doc/actions/runs/13381697432/job/37371236269

**Requirements check-list**

### DIFF
--- a/src/conf.py
+++ b/src/conf.py
@@ -134,6 +134,8 @@ intersphinx_mapping = {
 }
 
 nitpick_ignore = [
+    # external deps we don't cross-link to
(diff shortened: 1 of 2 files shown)


MetRonnie (approved):

oliver-sanders:
There's a new linkcheck error resulting from the deletion of cylc-flow tags (which were a convenient way to link the changelog for a release series without having to include the full version number). Will fix on this PR.

MetRonnie:
Oh no it's because I did that PR on master

PR #819, opened 2025-04-15
Note about platforms without Home directory

wxtim:
Response to [this comment on Element](https://matrix.to/#/!gMuCeOJBhYqgWDGwnR:matrix.org/$17446799571676fNkoK:matrix.org?via=matrix.org).

Hosted on: <internalserver>/doc/cylc/8.5.0.dev819/html/reference/config/writing-platform-configs.html#platform-with-no-home-directory

**Requirements check-list**

### DIFF
--- a/src/reference/config/writing-platform-configs.rst
+++ b/src/reference/config/writing-platform-configs.rst
@@ -339,7 +339,7 @@ Platform with no ``$HOME`` directory
 .. admonition:: Scenario
 
    You are trying to run jobs on a platform where the compute nodes don't
-   have a configured ``HOME`` directory.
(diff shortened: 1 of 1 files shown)


hjoliver:
If there's an expectation that platform hosts should see the same home directory, and a special workaround is required if that's not the case, that should be highlighted as in the main "what is a platform" documentation?

> What Are Platforms?
> 
> Platforms define settings, most importantly:
> -  A set of hosts.
> -  A job runner (formerly a batch system) where Cylc can submit a job.
> -  An install target for Cylc to install job files on.
> - **A common $HOME directory (if this is not the case, see below)**

PR #820, opened 2025-04-23
Redo xtrigger docs, with set xtriggers

hjoliver:
Goes with https://github.com/cylc/cylc-flow/pull/6695

However, the existing xtrigger docs were rubbish, so I rewrote them with the important concepts up front.

**Requirements check-list**

### DIFF
--- a/src/user-guide/writing-workflows/external-triggers.rst
+++ b/src/user-guide/writing-workflows/external-triggers.rst
@@ -1,47 +1,99 @@
 .. _Section External Triggers:
 
-External Triggers
-=================
+External Triggers (xtriggers)
+=============================
 
-External triggers allow tasks to trigger directly off of external events, which
-is often preferable to implementing long-running polling tasks in the workflow.
-The triggering mechanism described in this section is intended to replace the one
-one documented in :ref:`Old-Style External Triggers` (however, that one is a push
-mechanism, whereas this one involves regular polling by the scheduler).
+Xtriggers allow tasks to trigger off external conditions, by means of the
+scheduler periodically calling a Python function to check the condition.
 
-If you can write a Python function to check the status of an external
-condition or event, the :term:`scheduler` can call it at configurable
-intervals until it reports success, at which point dependent tasks can trigger
-and data returned by the function will be passed to the job environments of
-those tasks. Functions can be written for triggering off of almost anything,
-such as delivery of a new dataset, creation of a new entry in a database
-table, or appearance of new data availability notifications in a message
-broker.
+.. note::
+   Xtriggers should generally be preferred over the older push mechanism
+   described in :ref:`Old-Style External Triggers`. Xtriggers are more
+   transparent (they are exposed in the graph), easier to use (they don't
+   require modifying the external system), and efficient (only one call is
+   made, albeit repeatedly until success, for all dependent tasks).
 
-.. TODO - update this once we have static visualisation
+Cylc has several built-in xtriggers:
 
-   External triggers are visible in workflow visualizations as bare graph nodes (just
-   the trigger names). They are plotted against all dependent tasks, not in a
-   cycle point specific way like tasks. This is because external triggers may or
-   may not be cycle point (or even task name) specific - it depends on the
-   arguments passed to the corresponding trigger functions. For example, if an
-   external trigger does not depend on task name or cycle point it will only be
-   called once - albeit repeatedly until satisfied - for the entire workflow run,
-   after which the function result will be remembered for all dependent tasks
-   throughout the workflow run.
+- :ref:`Built-in Clock Triggers` - time trigger relative to cycle point
+- :ref:`Built-in Workflow State Triggers` - to trigger off tasks in other workflows
+- Several :ref:`Built-in Toy XTriggers` - to facilitate understanding of xtriggers
 
 .. TODO - auto-document these once we have a python endpoint for them
 
-Cylc has several built-in external trigger functions:
+Xtriggers are declared under :cylc:conf:`flow.cylc[scheduling][xtriggers]`
+by associating a short *label* with the function name and arguments to use
+in the workflow context. The label must be prefixed by the ``@`` character
+in graph triggers.
 
-- :ref:`Built-in Clock Triggers`
-- :ref:`Built-in Workflow State Triggers`
+Python function and installation requirements are described below in
+:ref:`Custom Trigger Functions`.
 
-Trigger functions are normal Python functions, with certain constraints as
-described below in :ref:`Custom Trigger Functions`.
+In the following example, the ``x1`` xtrigger represents the Python function
+``get_data``, with arguments as shown. The function will be called every 30
+seconds (the default interval is ``PT10S``). When it returns successs, the
+``process_data`` task will trigger.
 
-External triggers are configured in the
-:cylc:conf:`flow.cylc[scheduling][xtriggers]` section.
+.. code-block:: cylc
+
+   [scheduling]
+       [[xtriggers]]
+           x1 = get_data(loc="/path/to/data/source"):PT30S
+       [[graph]]
+           P1D = "@x1 => process_data => products"
+   [runtime]
+       [[get_data]]
+           # ...
+
+Xtrigger functions can return an arbitrary dictionary of information to be
+:term:`broadcast` to the job environment of dependent tasks. The environment
+variable names are constructed by prefixing the dictionary keys with the
+xtrigger label. For the example above:
+
+.. code-block:: python
+
+   # dictionary returned by get_data() on success:
+   {
+       "data_path": "/path/to/data",
+       "data_type": "netcdf"
+   }
+
+The job environment of the ``process_data`` task will then be given:
+
+.. code-block:: bash
+
+   # job environment of process_data:
+   x1_data_path="/path/to/data"
+   x1_data_type="netcdf"
+
+
+.. note::
+
+  You can manually satisfy xtriggers via the GUI or the command line (see
+  ``cylc set --help``) to allow dependent tasks to run even though the xtrigger
+  function did not return success.
+  For this to work without causing task failure, make sure you set default
+  values for xtrigger result variables in task scripting, e.g.:
+ 
+  .. code-block:: bash
+
+    # process_data task scripting
+    DATA="${x1_data_path:-}"
+    TYPE="${x1_data_type:-}"
+
+
+.. TODO - update this once we have static visualisation
+
+   External triggers are visible in workflow visualizations as bare graph nodes
+   (just the trigger names). They are plotted against all dependent tasks, not
+   in a cycle point specific way like tasks. This is because external triggers
+   may or may not be cycle point (or even task name) specific - it depends on
+   the arguments passed to the corresponding trigger functions. For example, if
+   an external trigger does not depend on task name or cycle point it will only
+   be called once - albeit repeatedly until satisfied - for the entire workflow
+   run, after which the function result will be remembered for all dependent
+   tasks throughout the workflow run.
+
 
 .. NOTE - from here on all references can start [xtriggers]
 
@@ -53,22 +105,24 @@ External triggers are configured in the
 Built-in Clock Triggers
 -----------------------
 
-These are more transparent (exposed in the graph) and efficient (shared among
-dependent tasks) than the older clock triggers described
-in :ref:`ClockTriggerTasks`.
+Clock xtriggers succeed when the real ("wall clock") time reaches some offset
+from the task's cycle point value.
 
-Clock triggers, unlike other trigger functions, are executed synchronously in
-the main process. The clock trigger function signature looks like this:
+.. note::
+
+   These should be used instead of the older task clock triggers documented in
+   :ref:`ClockTriggerTasks`.
+
+The clock xtrigger function signature looks like this:
 
 .. autofunction:: cylc.flow.xtriggers.wall_clock.wall_clock
 
-The ``offset`` argument is a datetime duration (``PT1H`` is 1
-hour) relative to the dependent task's cycle point (automatically passed to the
-function via a second argument not shown above).
+The ``offset`` argument is a datetime duration (e.g., ``PT1H`` is 1 hour)
+relative to the dependent task's cycle point.
 
-In the following workflow, task ``foo`` has a daily cycle point sequence,
-and each task instance can trigger once the wallclock time has passed its
-cycle point value by one hour:
+In the following workflow, task ``foo`` has a daily cycle point sequence, and
+each task instance will trigger when the real time is one hour past its cycle
+point value.
 
 .. code-block:: cylc
 
@@ -82,8 +136,8 @@ cycle point value by one hour:
        [[foo]]
            script = run-foo.sh
 
-Notice that the short label ``clock_1`` is used to represent the
-trigger function in the graph.
+Notice that the short label ``clock_1`` represents the trigger function in the
+graph.
 
 Argument keywords can be omitted if called in the right order, so the
 ``clock_1`` trigger can also be declared like this:
@@ -93,8 +147,7 @@ Argument keywords can be omitted if called in the right order, so the
    [[xtriggers]]
        clock_1 = wall_clock(PT1H)
 
-A zero-offset clock trigger does not need to be declared under
-the :cylc:conf:`[xtriggers]` section:
+A zero-offset clock trigger does not need to be declared under :cylc:conf:`[xtriggers]`:
 
 .. code-block:: cylc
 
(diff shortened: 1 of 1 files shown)


src/user-guide/writing-workflows/external-triggers.rst
```diff
@@ -53,22 +105,24 @@ External triggers are configured in the
...
+   :ref:`ClockTriggerTasks`.
+
+The clock xtrigger function signature looks like this:
 
 .. autofunction:: cylc.flow.xtriggers.wall_clock.wall_clock
 
-The ``offset`` argument is a datetime duration (``PT1H`` is 1
-hour) relative to the dependent task's cycle point (automatically passed to the
-function via a second argument not shown above).
+The ``offset`` argument is a datetime duration (e.g., ``PT1H`` is 1 hour)
```
wxtim, line 120: Does this add anything to content of the function doc? I'm not sure it's needed?

src/user-guide/writing-workflows/external-triggers.rst
```diff
@@ -82,8 +136,8 @@ cycle point value by one hour:
        [[foo]]
            script = run-foo.sh
 
-Notice that the short label ``clock_1`` is used to represent the
-trigger function in the graph.
+Notice that the short label ``clock_1`` represents the trigger function in the
+graph.
 
 Argument keywords can be omitted if called in the right order, so the
```
wxtim, line 142: Is this a necessary thing to say?

src/user-guide/writing-workflows/external-triggers.rst
```diff
@@ -82,8 +136,8 @@ cycle point value by one hour:
        [[foo]]
            script = run-foo.sh
 
-Notice that the short label ``clock_1`` is used to represent the
-trigger function in the graph.
+Notice that the short label ``clock_1`` represents the trigger function in the
+graph.
```
wxtim, line 140: Possible alt?
```suggestion
In the ``[xtriggers]`` section the trigger function is assigned the label ``clock_1``.
``@`` plus this label is what is then used in the graph.
```

src/user-guide/writing-workflows/external-triggers.rst
```diff
@@ -107,8 +160,7 @@ the :cylc:conf:`[xtriggers]` section:
        [[foo]]
            script = run-foo.sh
 
-However, when xtriggers are declared the name used must adhere to the following
-rules:
+However, xtrigger names must adhere to the following rules:
```
wxtim, line 163: "However" made me re-read the last example (starting at "A zero-offset..") but in reality, the two are unrelated. Consider removing the "However".

Perhaps consider putting this in a `..note::` directive?

src/user-guide/writing-workflows/external-triggers.rst
```diff
@@ -118,9 +170,14 @@ rules:
 Built-in Workflow State Triggers
 --------------------------------
 
-These can be used instead of the older workflow state polling tasks described
-in :ref:`WorkflowStatePolling` for inter-workflow triggering - i.e. to trigger local
-tasks off of remote task statuses or messages in other workflows.
+Workflow-state xtriggers succeed when a target task in another workflow achieves
+a given state or output.
```
wxtim, line 174: I think it's worth separating the general introduction from the explanation of why the terminology seems backwards (the target is the upstream dependency).

```suggestion
Workflow-state xtriggers succeed when a task in another workflow achieves
a given state or output.

The task in the other workflow is called the target, because it is the target of
the xtrigger.
```

src/user-guide/writing-workflows/external-triggers.rst
```diff
@@ -142,26 +199,22 @@ It must be installed and run under the name *up*, as referenced in the
 .. literalinclude:: ../../workflows/xtrigger/workflow_state/downstream/flow.cylc
```
wxtim, line 199: Any particular reason why these `flow.cylc` files are executable?

src/user-guide/writing-workflows/external-triggers.rst
```diff
@@ -142,26 +199,22 @@ It must be installed and run under the name *up*, as referenced in the
...
-The return value of the ``workflow_state`` trigger function looks like
-this:
+Try starting the downstream workflow first, then the upstream, and watch what
+happens. In each cycle point the ``@upstream`` trigger in the downstream workflow
+waits for the upstream task ``foo`` (with the same cycle point) workflow to generate
+the "data ready" message.
+
+Some important points to note:
+
+- The optional interval ``PT10S`` (10 seconds) determines how often the scheduler
```
wxtim, line 209: Is it worth having a warning here (or below this list) urgining users to consider what a sensible polling interval would be?

wxtim (commented):
Happy with most of this. Will label individual comments as blocking/non-blocking in terms of my review.

src/user-guide/writing-workflows/external-triggers.rst
hjoliver: >  from the explanation of why the terminology seems backwards 

I'm not sure I understand this comment. Did you omit some punctuation?

I prefer the original as I'm using "target" as merely an adjective. It's just saying we are targeting a particular task in another workflow - that's the task whose state we are interested. To me that's just normal descriptive English.  Your version sounds more like formalising "target" as special terminology, which wasn't my intention.

src/user-guide/writing-workflows/external-triggers.rst
hjoliver: Nope, news to me. I'll fix it.

src/user-guide/writing-workflows/external-triggers.rst
hjoliver: I'll remove it.

src/user-guide/writing-workflows/external-triggers.rst
hjoliver: I think it is necessary, to explain why you can call clock triggers the commonly-used shorthand way, but that for other xtriggers with more arguments, the order has to be right. It should be mentioned in the general section though.

src/user-guide/writing-workflows/external-triggers.rst
hjoliver: Huh, yeah, I suspect that line got repositioned at some point. It should also be in the general section, not clock triggers.

src/user-guide/writing-workflows/external-triggers.rst
hjoliver: I've moved this up to the general section too.

wxtim (approved):

hjoliver:
@wxtim - after reading through the xtrigger section in its entirety it became clear that a more extensive rewrite was desperately needed. That's nearly done but I haven't pushed the result up yet, so temporarily Drafting this.

hjoliver:
Ready to go again. The TOC now looks like this, with all the important general information at the top:

image

The `workflow-state` xtrigger documentation still needs a good rewrite, but I'll punt that to a new issue on cylc-flow.

src/user-guide/writing-workflows/external-triggers.rst
```diff
@@ -207,137 +451,127 @@ To see this, take a look at the job script for one of the downstream tasks:
...
-- They can be located either:
+Requirements
+^^^^^^^^^^^^
 
-  - In ``<workflow-dir>/lib/python/``;
-  - Anywhere in your ``$CYLC_PYTHONPATH``;
-  - Defined using the ``cylc.xtriggers`` entry point for an installed
-    package - see :ref:`developing.xtrigger.plugins`
+Xtrigger functions must be compatible with the Python version that runs the
+scheduler (see :ref:`Requirements` for the latest version specification).
```
wxtim, line 515: Worth adding a note about portability? 

Or adding a note to the document on [Portable workflows](https://github.com/cylc/cylc-doc/blob/master/src/workflow-design-guide/portable-workflows.rst) pointing out that just because you've deployed on Python 3.12 doesn't mean that Walruses and `f'{myvar=}` will work if someone else has Cylc deployed with Python 3.7?

src/user-guide/writing-workflows/external-triggers.rst
```diff
@@ -207,137 +451,127 @@ To see this, take a look at the job script for one of the downstream tasks:
...
-   because each call is executed in an independent process in the process
-   pool. If necessary the filesystem can be used for this purpose.
+where ``results`` is a flat (non-nested) dictionary of information to be passed
+to dependent tasks - see :ref:`Xtrigger Results`. Each dictionary key must be
+valid as an
+`environment variable <https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html>`_
+name.
 
+They can take arbitrary positional and keyword arguments, except that the
+``sequential`` keyword is reserved for :ref:`sequential xtriggers <Sequential Xtriggers>`.
```
wxtim, line 529:
```suggestion
They can take arbitrary positional and keyword arguments, except for the
keyword ``sequential``, which is reserved for :ref:`sequential xtriggers <Sequential Xtriggers>`.
```

src/user-guide/writing-workflows/external-triggers.rst
```diff
@@ -207,137 +451,127 @@ To see this, take a look at the job script for one of the downstream tasks:
...
 
-.. spelling:word-list::
+They cannot store data Pythonically between invocations, because each call
+is executed via a wrapper in a new subprocess. If necessary the filesystem
+could be used for persistent data.
```
wxtim, line 533: Is this _really_ a pattern we'd recommend?

src/user-guide/writing-workflows/external-triggers.rst
```diff
@@ -207,137 +451,127 @@ To see this, take a look at the job script for one of the downstream tasks:
...
-  take too long to return, the process will be killed
-- they are shared for efficiency: a single call will be made for all
-  triggers that share the same function signature - i.e.\ the same function
-  name and arguments
-- their return status and results are stored in the workflow DB and persist across
-  workflow restarts
-- their stdout, if any, is redirected to stderr and will be visible in
-  the workflow log in debug mode (stdout is needed to communicate return values
-  from the sub-process in which the function executes)
+   vv
```
wxtim, line 567: ?

src/user-guide/writing-workflows/external-triggers.rst
```diff
@@ -207,137 +451,127 @@ To see this, take a look at the job script for one of the downstream tasks:
...
+to dependent tasks - see :ref:`Xtrigger Results`. Each dictionary key must be
+valid as an
+`environment variable <https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html>`_
+name.
 
+They can take arbitrary positional and keyword arguments, except that the
+``sequential`` keyword is reserved for :ref:`sequential xtriggers <Sequential Xtriggers>`.
 
-.. spelling:word-list::
+They cannot store data Pythonically between invocations, because each call
```
wxtim, line 531: [Don't really care that much, arguable style point] IMO I think "Custom xtrigger functions" or "custom xtriggers", or even just "xtriggers" would be more desirable than "they".

wxtim (approved):
I've made some comments, but there's nothing which should block merging.

For other MO reviewers, this is built on my internal server at `/doc/cylc/8.4.3.dev820/html/user-guide/writing-workflows/external-triggers.html#`

Some style guides would pick you up on use of passive voice, and I think that's worth considering, because it makes the user think about the scheduler and the xtrigger functions as actors doing something, which may help in terms of general Cylc understanding. I think that the following is clearer where we are assigning responsibility to the scheduler:

> Future tasks that depend on the same xtrigger will be satisfied immediately without calling the function again
vs
> The scheduler marks future tasks depending on the same xtrigger as satisfied without calling the xtrigger function again.

src/user-guide/writing-workflows/external-triggers.rst
hjoliver: Yep, I prefer your version.

src/user-guide/writing-workflows/external-triggers.rst
hjoliver: Yes, I was unsure of this myself. I decided a bulleted list of long sentences doesn't look great in this context (it's a more of a book than a powerpoint presentation), but repeating the object on every line seemed too much.

src/user-guide/writing-workflows/external-triggers.rst
```diff
@@ -353,96 +587,43 @@ The :py:mod:`cylc.flow.xtriggers.xrandom` xtrigger module contains an example
...
-- aside from the predefined zero-offset ``wall_clock`` trigger, all
-  unique trigger function calls must be declared *with all of
-  their arguments* under the :cylc:conf:`[xtriggers]` section, and
-  referred to by label alone in the graph. It would be convenient (and less
-  verbose, although no more functional) if we could just declare a label
-  against the *common* arguments, and give remaining arguments (such as
-  different wallclock offsets in clock triggers) as needed in the graph.
-- we may move away from the string templating method for providing workflow
-  and task attributes to trigger function arguments.
+- Trigger names cannot be used in conditional (OR) expressions in the graph;
```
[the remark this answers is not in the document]
hjoliver: No that you say it, I agree. I'm not sure how this section ended up in the User Guide. I'll just remove it, to a cylc-flow Issue...

src/user-guide/writing-workflows/external-triggers.rst
hjoliver: OK I've gone with repetition, as the lesser of two evils.

hjoliver:
I agree with using active voice where possible, and sometimes I'm sure I do use passive voice on auto-pilot when I shouldn't, but sometimes it's appropriate.

>  Future tasks that depend on the same xtrigger will be satisfied immediately without calling the function again
> 
> vs
> 
>  The scheduler marks future tasks depending on the same xtrigger as satisfied without calling the xtrigger function again.

In this case, the passive version (as written at least) is more succinct, and I think the "actor" should be irrelevant to the reader - it matters that the xtrigger will be satisfied, but it doesn't matter (or it's obvious) that the "the scheduler" will be what does that.

hjoliver:
All done.

hjoliver:
(Last commit replaces a link to a wikipedia page that got removed yesterday).

hjoliver:
All done. Last commit replaces a link to a wikipedia page that got removed yesterday.

wxtim:
> Going deep on https://www.grammarly.com/blog/grammar/passive-voice/ (because I love this stuff 😁 )

I did before I made the comment. I like [this](https://advice.writing.utoronto.ca/revising/passive-voice/) because it has a checklist.

oliver-sanders (commented):
Raised some edits as a PR: https://github.com/hjoliver/cylc-doc/pull/4

src/user-guide/writing-workflows/external-triggers.rst
```diff
@@ -1,47 +1,318 @@
...
+
+Xtriggers allow tasks to trigger off of arbitrary external conditions. The
+:term:`scheduler` periodically calls a Python function, to check on the
+condition, until it returns success to satisfy dependent tasks.
+
+
+.. note::
+   Compared to the older :ref:`Old-Style External Triggers`, xtriggers are
+   visible in the graph, they are shared efficiently by all dependent tasks,
+   and they are entirely contained within the workflow.
```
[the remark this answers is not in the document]
hjoliver: I think I avoided saying that the new xtriggers are entirely superior, because the old ones are in principle better if your use case can easily support a push mechanism. The new ones are technically superior in every other way though (written in the graph, no need to modify external systems).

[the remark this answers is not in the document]
hjoliver: > I think you mean they are written in the [graph] as opposed to visible in the graph view.

Yes, I suppose I should say "visible in the graph configuration", for now.

hjoliver:
Having rendered the docs I'm happy with your correcting heading nesting @oliver-sanders 

I still think we should have some text that recommends xtriggers of ext-triggers for most uses, because the fact is they are much easier to use for most use cases, and new users might waste time figuring that out for themselves if we don't tell them.  Reasons: https://github.com/hjoliver/cylc-doc/pull/4#discussion_r2229708606 

(I do agree that push is superior to polling in principle, but in practice the polling solution is self-contained and relatively easy, and the downsides of polling *often* don't matter at all).

oliver-sanders:
>  It's just that until then, it's not very helpful to tell users to use ext-triggers for most of what they current do with xtriggers - it just isn't practical.

It isn't practical to tell users to use ext-triggers to solve polling problems.

However, it also isn't practical to tell users to xtriggers to solve event-driven problems.

Don't assume it's hard to implement a push pattern, it can be very easy, it all depends on the use case.

PR #798, opened 2025-02-18
Disambiguate 'active tasks'.

hjoliver:
We're still using "active tasks" to mean both:
  1. tasks in the active window, and
  2. submitted or running tasks, specifically

This PR ensures that *"tasks in the active window"* is used in the first case, and clarifies the definitions in both cases.

**Requirements check-list**
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.

PR #824, opened 2025-05-08
Disambiguate the word "active".

hjoliver:
(This branch is a few months old; it supersedes #798 which tried to be lighter weight but was still confusing).

Our overloading of the term "active" is still confusing:
- *"active tasks"* typically means tasks with active jobs
   - (for historical reasons that don't really matter anymore)
- the *"active window"* (of the workflow) is the scheduler task pool 
   - (task pool is no longer a user-facing term)
   - the active window contains many tasks that are "not active"!
- the *`n=0` window* is identical to the "active window"
   - (the latter term predates the former, but we kept both for reasons that don't really matter anymore)
- we now use the terms "active" and "inactive" internally, in scheduler code, to distinguish pool and future tasks

**This PR resolves the problem by dropping the redundant term "active window" and redefining "active task" slightly:**
- an **"active task"** is a task under active management by the scheduler (i.e. in the task pool).
- active tasks comprise the **n=0 window** of the workflow

Rationale:
- with task/job separation we don't need a special term for a *task* with an active job
- the active window *should* contain just the active tasks, by any sensible definition of the terms
- but we have to use "n=0 window" for other reasons anyway, so we might as well drop the redundant term
  - (however, this change makes "active window" perfectly compatible with "active tasks" and "n=0 window", so no biggie, and not confusing anymore, if we still use the term accidentally or otherwise)

**Requirements check-list**

### DIFF
--- a/src/glossary.rst
+++ b/src/glossary.rst
@@ -68,50 +68,36 @@ Glossary
       Submit number also appears in the job log path so that job log files
       don't get overwritten.
 
-
-   active
    active task
-      An active task is a task which is near ready to run, in the process of
-      running, or which requires user intervention. These are all the tasks
-      being actively managed by the scheduler at this point in the run. 
+      Active tasks are those tasks currently held in working memory to feed
+      the scheduling algorithm. They comprise ``n=0`` basis of the GUI
+      :term:`n-window`, so they are always visible in the GUI.
 
-      Active tasks are:
+      Active tasks include:
 
-      - Tasks which have some, but not all of their
-        :term:`prerequisites <prerequisite>` satisfied.
-      - ``waiting`` tasks, that are actively waiting on:
+      - ``submitted`` and ``running``, with active jobs)
+      - ``preparing`` tasks in the job submission pipeline
+      - ``waiting`` tasks that are nearly ready to run but:
 
-        - :term:`xtriggers <xtrigger>`.
-        - :ref:`internal queues <InternalQueues>`
-        - :ref:`runahead limit <RunaheadLimit>`
+        - have partially satisfied :term:`prerequisites <prerequisite>`
+        - or are waiting on :term:`xtriggers <xtrigger>`,
+          :ref:`internal queues <InternalQueues>`, or the :ref:`runahead limit <RunaheadLimit>`
 
-      - ``preparing`` tasks - i.e. tasks in the process of submitting jobs
-      - ``submitted`` and ``running`` tasks - i.e. those with active jobs
       - tasks that reached a :term:`final status` without completing their
         :term:`required outputs <required output>`
         (e.g. a task failed where success was required).
 
-      Active tasks are in the ``n=0`` :term:`window <n-window>` which means they
-      will always be displayed in the GUI and Tui.
 
-      The distinction between active and non-active tasks is important for
-      the computing of the :term:`runahead limit`.
-
+   n-window
+      The GUI provides a view of the workflow extending ``n`` graph edges out
+      from :term:`active tasks <active task>` - which comprise the ``n=0``
+      window. The default n-window extent is ``n=1``. 
 
    active cycle
-      A cycle point is active if it contains any :term:`active tasks <active task>`.
-
-      Active cycles are counted towards the :term:`runahead limit`.
-
-
-   window
-   n-window
-   active window
-      The GUI provides a :term:`graph`-based window or view of the workflow at
-      runtime, including tasks out to ``n`` graph edges from current
+      A cycle point is considered to be active if it contains any
       :term:`active tasks <active task>`.
 
-      Active tasks form the ``n=0`` window.
+     Active cycles count toward the :term:`runahead limit`.
 
       .. seealso::
 
@@ -381,6 +367,9 @@ Glossary
       ``3``, etc.
       For :term:`datetime cycling` they will be :term:`ISO 8601` datetimes,
       e.g. ``2000-01-01T00:00Z``.
+
+      Cylc has no global cycle loop, so each task instance has its own cycle
+      point label.
 
       .. seealso::
 
@@ -993,7 +982,7 @@ Glossary
       This refers to starting a new instance of the Cylc :term:`scheduler`
       program to manage a particular :term:`workflow`. This can be from
       scratch, for installed workflows that haven't run previously, or to
-      restart one that shut down prior to :term:`completion <workflow completion>`.
+      restart one that shut down prior to :ref:`completion <workflow completion>`.
 
       .. seealso::
 
@@ -1170,7 +1159,7 @@ Glossary
    stop
    shutdown
       A :term:`scheduler` can shut down on request, or automatically on
-      :term:`workflow completion`. The :term:`workflow` is then stopped and no
+      :ref:`workflow completion`. The :term:`workflow` is then stopped and no
       further :term:`jobs <job>` will be submitted.
 
       By default, the scheduler waits for any submitted or running task
@@ -1216,9 +1205,10 @@ Glossary
       workflow :term:`source directory` before reload, rather than made by
       editing the installed files directly.
 
-      :ref:`RemoteInit` will be redone for each job platform, when the first job is submitted there after a reload.
+      :ref:`RemoteInit` will be redone for each job platform, when the first
+      job is submitted there after a reload.
 
-      Any :term:`task` that is :term:`active <active task>` at reload
+      Any task that is :term:`active <active task>` at reload
       will continue with its pre-reload configuration.
       The next instance of the task (at the next cycle point)
       will adopt the new configuration.
@@ -1433,7 +1423,7 @@ Glossary
       - Or, if expiry is optional, then the outputs are complete if it expires.
 
       Tasks that achieve a :term:`final status` with complete outputs have done
-      their job, allowing the workflow to move on.
+      their job in the workflow, allowing the scheduler to move on.
 
       Tasks that achieve a final status with incomplete outputs are retained in
       :term:`n=0 <n-window>` pending user intervention, and will :term:`stall`
@@ -1587,7 +1577,7 @@ Glossary
 
    stall
    stalled workflow
-      A stalled workflow has not run to :term:`completion <workflow completion>`
+      A stalled workflow has not :ref:`run to completion <workflow completion>`
       but cannot continue without manual intervention. 
 
       A stalled scheduler stays alive for a configurable timeout period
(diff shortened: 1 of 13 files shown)


src/glossary.rst
```diff
@@ -68,50 +68,36 @@ Glossary
...
 
-      Active tasks are in the ``n=0`` :term:`window <n-window>` which means they
-      will always be displayed in the GUI and Tui.
-
-      The distinction between active and non-active tasks is important for
-      the computing of the :term:`runahead limit`.
 
+   n-window
+      The GUI provides a view of the workflow extending ``n`` graph edges out
+      from :term:`active tasks <active task>` - which comprise the ``n=0``
```
oliver-sanders, line 93: Comprise is a good word for the job, but "form" might be easier for users to understand:

```suggestion
      from :term:`active tasks <active task>` - which form the ``n=0``
```

src/user-guide/writing-workflows/scheduling.rst
```diff
@@ -2219,15 +2219,12 @@ executing, e.g:
...
+configurable limit.
 
-The runahead limit is defined as an interval measured from the oldest active cycle.
-A cycle is considered to be "active" if it contains any :term:`active` tasks
-(e.g. running tasks).
-
-Tasks in cycles which are beyond the limit are called :term:`runahead` tasks
-and are displayed in the GUI/Tui with small circle above them:
+Tasks in cycles beyond the runahead limit are called "runahead limited" and are
+displayed in the GUI/Tui with small circle above them:
```
oliver-sanders, line 2227: Sadly this is not true, see https://github.com/cylc/cylc-flow/issues/5580

oliver-sanders (commented):
LGTM, will need a quick skim of `cylc.flow.cfgspec` as there are a few mentions of "active" there too.

src/glossary.rst
```diff
@@ -68,50 +68,36 @@ Glossary
...
 
-
-   active
    active task
-      An active task is a task which is near ready to run, in the process of
-      running, or which requires user intervention. These are all the tasks
-      being actively managed by the scheduler at this point in the run. 
+      Active tasks are those tasks currently held in working memory to feed
+      the scheduling algorithm. They comprise ``n=0`` basis of the GUI
+      :term:`n-window`, so they are always visible in the GUI.
```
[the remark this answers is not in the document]
hjoliver: Yes, it should.

src/glossary.rst
```diff
@@ -68,50 +68,36 @@ Glossary
...
 
-      Active tasks are:
+      Active tasks include:
 
-      - Tasks which have some, but not all of their
-        :term:`prerequisites <prerequisite>` satisfied.
-      - ``waiting`` tasks, that are actively waiting on:
+      - ``submitted`` and ``running``, with active jobs)
```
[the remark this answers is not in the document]
hjoliver: (Yes, I've gone with "task with active jobs" on the companion cylc-flow doc PR.)

src/user-guide/writing-workflows/scheduling.rst
hjoliver: Ah, yeah, I did know that, I was just tweaking the existing text on autopilot there. I'll rephrase it.

hjoliver:
> will need a quick skim of cylc.flow.cfgspec as there are a few mentions of "active" there too.

See https://github.com/cylc/cylc-flow/pull/6618 from Feb 19 for the PR that this superseded. I've now finished it off and undrafted it.

src/user-guide/writing-workflows/scheduling.rst
hjoliver: Now this:
```
Runahead limiting confines workflow activity to a configurable number of cycles
beyond the earliest :term:`active cycle`.

Tasks in the :term:`n=0 window <n-window>` at the runahead limit are actively
held back, and are displayed in the GUI/Tui with a small circle above them.

.. note::

   Tasks in the :term:`n=1 window <n-window>` are not displayed as runahead
   limited; they just form the future graph and are not yet being actively limited.
```

hjoliver:
Build requires https://github.com/cylc/cylc-flow/pull/6618

oliver-sanders (approved):

src/glossary.rst
```diff
@@ -68,50 +68,36 @@ Glossary
...
       don't get overwritten.
 
-
-   active
    active task
-      An active task is a task which is near ready to run, in the process of
-      running, or which requires user intervention. These are all the tasks
-      being actively managed by the scheduler at this point in the run. 
+      Active tasks are those tasks currently held in working memory to feed
+      the scheduling algorithm. They form the ``n=0`` :term:`n-window`,
```
wxtim, line 73: Might simplify the sentence without losing much?
```suggestion
      Active tasks are those tasks held in memory to feed
      the scheduling algorithm. They form the ``n=0`` :term:`n-window`,
```

src/user-guide/running-workflows/tasks-jobs-ui.rst
```diff
@@ -104,31 +104,26 @@ The "n" Window
...
-:term:`cycling workflows <cycling workflow>` with no :term:`final cycle point`.
+Cylc workflow :term:`graphs <graph>` can be very large, or infinite in
+extent for :term:`cycling workflows <cycling workflow>` with no
+:term:`final cycle point`.
 
 Consequently the GUI often can't display "all of the tasks" at once. Instead
-it displays all :term:`active tasks <active task>` (e.g. running tasks)
-as well as any tasks which are a configurable number of tasks away from
-them in the task dependency :term:`graph`.
+it displays all tasks a configurable :term:`n-window` around the current
```
wxtim, line 112:
```suggestion
it displays all tasks in a configurable :term:`n-window` around the current
```

src/user-guide/running-workflows/tasks-jobs-ui.rst
```diff
@@ -104,31 +104,26 @@ The "n" Window
...
    :align: center
 
 
 n=0:
-   The ``n=0`` window contains :term:`active tasks <active task>`. An active
-   task is a task which is near ready to run, in the process of running, or
-   which requires user intervention (see the :term:`glossary <active task>`
-   for a more detailed description).
+   The ``n=0`` window contains current :term:`active tasks <active task>`: those
+   that are near ready to run, running, or which require user intervention.
```
wxtim, line 121: [picky]

```suggestion
   that are near ready to run, running, or which may require user intervention.
```

src/user-guide/running-workflows/workflow-completion.rst
```diff
@@ -3,43 +3,37 @@
...
 
 Scheduler Stall
 ===============
 
-A stalled workflow has not run to :term:`completion <workflow completion>`
-but cannot continue without manual intervention. 
+If there are no tasks waiting on as yet unsatisfied external constraints
+such clock and xtriggers, and all activity has ceased but workflow has
+not :ref:`run to completion <workflow completion>`, then it
+has stalled and requires manual intervention to continue.
```
wxtim, line 18: Hemmingway editor gives this a reading age of "Post-graduate". This suggestion is US grade 7 (13-14 YO).

```suggestion
A workflow has stalled if:

* No tasks are waiting on unstatisfied external events, like clock triggers and xtriggers.
* AND All activity has ceased.
* AND The workflow has not run to completion.

A workflow which has stalled requires manual intervention to continue.
```

src/user-guide/running-workflows/workflow-completion.rst
```diff
@@ -3,43 +3,37 @@
...
-or otherwise) it will remain in the stalled state on restart.
+to allow you to intervene, e.g. by manually triggering an incomplete
+task after fixing the bug that caused it to fail.
 
-Stalls are often caused by unexpected task failures, either directly (tasks
-with :term:`final status` and :term:`incomplete outputs <output completion>`)
-or indirectly (tasks with partially satisfied :term:`prerequisites <prerequisite>`,
-downstream of an unexpected failure).
+If a stalled workflow does eventually shut down, on the stall timeout
+or by stop command, it will immediately stall again on restart.
```
wxtim, line 31: Worth saying that you can get around this by using `cylc play --pause`?

src/user-guide/running-workflows/workflow-completion.rst
```diff
@@ -3,43 +3,37 @@
...
 
 .. warning::
 
-   At present you have to consult the :term:`scheduler log` to see the reason
-   for a stall.
-
+   At present you have to look at the :term:`scheduler log` to see
```
wxtim, line 35: [picky] At present suggests that this might be different in the past or future. Why not just go for

> Look in the scheduler log to see which tasks caused a stall.

src/user-guide/writing-workflows/runtime.rst
```diff
@@ -837,10 +840,10 @@ Late Events
...
 .. warning::
 
-  The scheduler can only check for lateness once a task has appeared in its
-  active task window. In Cylc 8 this is usually when the task is actually
-  ready to run, which severely limits the usefulness of late events as
-  currently implemented.
+  The scheduler can only check for lateness once a task becomes
+  :term:`active <active task>`. In Cylc 8 this usually means the task
+  is ready, or nearly ready, to run, which limits the usefulness of late
+  events as currently implemented.
```
wxtim, line 846: Do we need the "as currently implemented"?

wxtim (approved):
No dealbreakers here.

src/user-guide/running-workflows/workflow-completion.rst
hjoliver: No, that's what we used to have to do (pause, because it would shutdown immediately instead of going to the stalled state as it does now).

src/user-guide/running-workflows/workflow-completion.rst
hjoliver: I think the reasoning there is to suggest that information will be exposed in the UIs in the future.  However, I agree, best to just document that current situation, generally. So I'll take your suggestion.

src/user-guide/writing-workflows/runtime.rst
hjoliver: You're right, we don't.

src/glossary.rst
```diff
@@ -382,6 +368,9 @@ Glossary
       For :term:`datetime cycling` they will be :term:`ISO 8601` datetimes,
       e.g. ``2000-01-01T00:00Z``.
 
+      Cylc has no global cycle loop, so each task instance has its own cycle
```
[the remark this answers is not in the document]
hjoliver: This is the "no barrier between cycles" thing - I'll make that more explicit.

src/user-guide/running-workflows/workflow-completion.rst
hjoliver: I'll add "to await manual intervention" at the end of the sentence.

src/glossary.rst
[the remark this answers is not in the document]
hjoliver: Changed to this:
```
Cylc can run multiple cycles at once, dependencies allowing, so each
task instance has its own cycle point label.
```

PR #351, opened 2021-11-30
Doc overhaul 2

hjoliver:
Supersedes #331 

This is the 2nd (or 3rd?) large and last PR to do a mass doc upgrade for Cylc 8, for the beta-3 release.

I have attempted to go through (almost) all of the documentation and:
- update or remove Cylc 7 docs that are now outdated, e.g. suicide triggers
- add new sections for new Cylc 8 features still missing, e.g. reflow and flow numbers, optional and expected outputs, incomplete tasks

While reading through I did a fair amount of minor revision, in an attempt to make a lot of stuff more clear and concise.

This is not intended to be the final solution!!  Just a bulk upgrade to end up with documentation that a) does not document obsolete Cylc 7 features; and b) does document all major new Cylc 8 features. I.e. to create documentation that is (mostly) not actually **wrong**.

For future improvements, note also #350 

Reviewers: feel free to highlight minor errors etc. that I can fix quickly. Any major or structural can be done in future, after the release is announced!

PR #830, opened 2025-05-23
Remove an inaccurate sentence.

hjoliver:
image

Artificial dependency, as defined in the glossary:

> An artificial [dependency](https://cylc.github.io/cylc-doc/stable/html/glossary.html#term-dependency) in the graph does not reflect real dependence between the tasks involved. This can sometimes be useful but should be avoided if possible. Artificial dependencies muddy the real dependencies of the workflow and they may unnecessarily constrain the scheduler.

This dependency `c | r:fail => d` is not artificial because `d` presumably does depend on real outputs of the upstream tasks. The fact that either one of the branches can generate the outputs doesn't make the dependency artificial.

`git blame` fingers me for this, although the commit name suggests I was combining existing doc sections, so I'm not sure if there is an original author out there who might disagree with this change.

**Requirements check-list**

### DIFF
--- a/src/user-guide/writing-workflows/scheduling.rst
+++ b/src/user-guide/writing-workflows/scheduling.rst
@@ -1935,8 +1935,7 @@ The ``?`` symbol denotes an :term:`optional output` which allows the graph to
 branch.
 
 Note the last line of the graph ``c | r => d`` allows the graph to
-continue on to ``d`` regardless of the path taken. This is an :term:`artificial
-dependency`.
+continue on to ``d`` regardless of the path taken.
 
 This is a simple example of a "switch" pattern, the task ``b`` being the switch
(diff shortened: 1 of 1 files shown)


oliver-sanders:
Looks like this "artificial dependency" reference came from https://github.com/cylc/cylc-doc/pull/351 ([patch](https://github.com/cylc/cylc-doc/pull/351/files#diff-5db318277500dec9b8bac896c33c59c300f4f8270e419e51f907a9b376a56e11R1790)). I think it was accidentally transferred from the "custom outputs" example (where we needed an artificial dependency to avert premature shutdown), This has since been [re-written to use completion expressions](https://cylc.github.io/cylc-doc/stable/html/user-guide/writing-workflows/scheduling.html#completion-expressions).

oliver-sanders (approved):

PR #848, opened 2025-07-15
deprecate compatibility mode

oliver-sanders:
* Addresses https://github.com/cylc/cylc-flow/issues/6849
* Goes with https://github.com/cylc/cylc-flow/pull/6853

**Requirements check-list**

### DIFF
--- a/src/7-to-8/caveats.rst
+++ b/src/7-to-8/caveats.rst
@@ -5,14 +5,6 @@ There are a few Cylc 7 features which do not yet have direct replacements in
 Cylc 8. These features will be re-implemented in future releases.
 
 
-Cylc Flow
----------
-
-Multiple Flows
-   The new :term:`scheduler` can manage multiple flows in the workflow graph.
-   Commands and options for interacting with flows are still being refined.
-
-
 Browser Based UI
 ----------------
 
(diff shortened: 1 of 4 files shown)


src/7-to-8/caveats.rst
```diff
@@ -5,14 +5,6 @@ There are a few Cylc 7 features which do not yet have direct replacements in
...
 
 
-Cylc Flow
----------
-
-Multiple Flows
-   The new :term:`scheduler` can manage multiple flows in the workflow graph.
-   Commands and options for interacting with flows are still being refined.
-
-
```
oliver-sanders, line 15: Off topic, but I think this can be removed now that flows are supported by:

* stop
* trigger
* set
* remove

src/7-to-8/caveats.rst
```diff
@@ -39,14 +31,6 @@ Installing Workflows
...
 Rose Edit
    Rose Edit is awaiting reimplementation.
-Xtrigger Visibility
-   Xtriggers are not yet visible in the UI.
-
-   * https://github.com/cylc/cylc-ui/issues/331
-Documentation / Orientation Guide
-   Some form of documentation will be provided within the UI itself.
-
-   * https://github.com/cylc/cylc-ui/issues/155
```
oliver-sanders, line 49: Should probs stop saying this?

hjoliver (approved):
Good, one review will do for this. Double checked with a quick grep for "compat".

PR #851, opened 2025-07-29
skip mode modifier

wxtim:
Companion to https://github.com/cylc/cylc-ui/issues/1949

Built at https://localserver/usename/doc/cylc/8.4.5.dev851/html/index.html

**Requirements check-list**


src/glossary.rst
```diff
@@ -1759,13 +1759,15 @@ Glossary
          * :ref:`Runahead Limiting`
          * :term:`active cycle`
 
+   skip mode
+      A task run mode that allows setting tasks to run instantly.
```
hjoliver, line 1763: They don't "run instantly" so much as skip running but complete outputs. Not sure how best to put that succinctly. Maybe this?:

```suggestion
      A task run mode that skips execution but completes outputs as if the task had run.
```

hjoliver (approved):

hjoliver:
> Built at https://localserver/usename/doc/cylc/8.4.5.dev851/html/index.html

(What is that URL?)

One review will do as the icon has essentially been approved on the issue already.

wxtim:
@hjoliver 

> (What is that URL?)

Unhelpful for you. Sorry. 

We have an internal server which will serve a folder from everyone's linux account `${HOME}` directory and I have my docs builds symlinked there. People who work in my org can replace `localserver/username` with a predictable value to see the built docs. It's probably paranoia, but I think I should keep internal server details hidden.

Issue #779, opened 2024-11-25
Undocumented `$CYLC_WORKFLOW_LOG_DIR`


MetRonnie:
`CYLC_WORKFLOW_LOG_DIR` needs adding to the job script environment variables reference: https://cylc.github.io/cylc-doc/latest/html/reference/job-script-vars/index.html

A quirk is that this points to `${CYLC_WORKFLOW_RUN_DIR}/log/scheduler` rather than just `${$CYLC_WORKFLOW_RUN_DIR}/log` - see https://github.com/cylc/cylc-flow/issues/6494

Also, for its Jinja2 equivalent, if we don't change where it points to, we need to update the description here to reflect this quirk: https://github.com/cylc/cylc-doc/blob/d76834ff426e94d590987058c05d5e315bdc9b2a/src/user-guide/writing-workflows/jinja2.rst?plain=1#L156-L162

oliver-sanders:
(when addressing this issue, please go through all the env vars exported in the job.sh file to check there aren't other omissions)

PR #853, opened 2025-08-01
Improve Job Script Environment Variables reference

jfrost-mo:
The variables `CYLC_ENV_NAME` and `CYLC_WORKFLOW_LOG_DIR` were newly added, and existing variables had their descriptions and examples improved, notably adding the run sub-directory to the example paths.

The longer name for `CYLC_TASK_LOG_ROOT` does however make a scrollbar appear. To counter this, and also to make better use of the semantics of HTML, I've then converted this reference to a definitions list. This should be more accessible, easier to copy from, and IMO is easier to read.

Rendered version: https://tmpweb.net/V5cucaJ-1HKF/reference/job-script-vars/

Fixes: #779

**Requirements check-list**


jfrost-mo:
I realise I forgot to copy the second commit message into the PR description. The first commit adds a number of missing variables and improves descriptions, while the second changes the formatting to use a HTML definition lists.

Would you like me to convert this to use the envvar domain?

src/reference/job-script-vars/var-list.rst
```diff
@@ -0,0 +1,119 @@
+CYLC_VERSION
+    Version of cylc installation used
+
+CYLC_ENV_NAME
+    Environment from which cylc commands run
+
```
oliver-sanders, line 6: Suggest leaving this one undocumented:

```suggestion
```

Reason being:

* This variable may, or may not be set depending on how Cylc is deployed because it is set by the wrapper script, not by Cylc itself.
* It's probably a bad idea to use this in your workflows (i.e. don't write your tasks into the Cylc environment).

jfrost-mo:
Spelling fixed in c2474b50a55eac7ccca0c07ba68766b1c4dd4039.

jfrost-mo:
Rebased to fix conflict. (And add gregorian to non-proper nouns, due to how we are using it.)

PR #754, opened 2024-07-16
workaround build errors with sphinx 7.4.4

oliver-sanders:
**Requirements check-list**
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.

PR #855, opened 2025-08-05
Remove upper pin on Sphinx version

MetRonnie:
Upper pin was introduced in #754, not sure what specifically for. But in Python 3.13+, need newer versions to avoid https://github.com/sphinx-doc/sphinx/issues/13178

**Requirements check-list**


oliver-sanders (approved):
Happy as long as it works!

oliver-sanders (approved):

PR #865, opened 2025-08-27
Add sphinx copybutton extension

wxtim:
**Requirements check-list**

Add a copy button to code snippets.

image

I've wanted to do this for ages, but last time I looked this extension didn't exist. 

Built at https://████████████████/doc/cylc/8.5.2.dev/html/tutorial/scheduling/graphing.html#example

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -52,6 +52,7 @@ install_requires =
     hieroglyph>=2.1.0
     setuptools>=50
     sphinx>=7.1, !=7.4.4
+    sphinx-copybutton
     sphinx-design>=0.5.0
     sphinx_rtd_theme>=1.0.0
     sphinxcontrib-svg2pdfconverter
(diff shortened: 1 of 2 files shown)


oliver-sanders:
(please test this out with `diff` code-blocks)

MetRonnie (approved):
Works well apart from diff blocks:

Image

What ends up copied:
```
 [scheduler]
    allow implicit tasks = True
 [scheduling]
    cycling mode = integer
    initial cycle point = 1
    [[graph]]
-        R1 = """
        P1 = """
            buy_ingredients => make_dough
            pre_heat_oven & make_dough => bake_bread
            bake_bread => sell_bread & clean_oven
        """
```

However I'm not sure if there is a good need to copy diff blocks.

Also, it might be tricky to solve as there is something funky going on with diff blocks when you try to select the text:

image

oliver-sanders:
> Also, it might be tricky to solve as there is something funky going on with diff blocks:

Diff blocks have some CSS rules to make it easier for users to copy paste:

https://cylc.github.io/cylc-sphinx-extensions/extensions/cylc.sphinx_ext.diff_selection.html

However, this is CSS only, just display, nothing is different about these blocks otherwise.

Suggest turning the copy feature off for diff blocks to avoid user pain in training sessions? Or finding a way to hack it?

PR #866, opened 2025-08-27
Minimum Python 3.12

MetRonnie:
Accompanies https://github.com/cylc/cylc-flow/pull/6809

**Requirements check-list**
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.

PR #867, opened 2025-08-27
Command task matching

oliver-sanders:
**Built On:** https://github.com/cylc/cylc-doc/pull/866

**Requirements check-list**

### DIFF
--- a/src/reference/changes.rst
+++ b/src/reference/changes.rst
@@ -36,6 +36,77 @@ Cylc 8.6
    :cylc-rose: `1.5 <https://github.com/cylc/cylc-rose/blob/master/CHANGES.md>`__
    :rose: `2.4 <https://github.com/metomi/rose/blob/master/CHANGES.md>`__
 
+
+Task Matching
+^^^^^^^^^^^^^
+
+Various Cylc commands (e.g. ``cylc trigger``, ``cylc set`` and ``cylc hold``)
+allow us to select the tasks we want to operate on using cycle points,
+family names or globs, e.g:
+
+.. code-block:: bash
+
+   # trigger active tasks in cycle "2000"
+   cylc trigger workflow//2000
+
+   # trigger active tasks in "FAMILY" in cycle "2000"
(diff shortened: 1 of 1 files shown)


src/reference/changes.rst
```diff
@@ -36,6 +36,95 @@ Cylc 8.5
...
+   # trigger active tasks in cycle "2000"
+   cylc trigger workflow//2000
+
+   # trigger active tasks in "FAMILY" in cycle "2000"
+   cylc trigger workflow//2000/FAMILY
+
+   # trigger all tasks in cycle "2000" with names starting with "foo"
+   cylc trigger workflow//2000/foo*
+
+However, Cylc would only operate on :term:`active tasks <active task>`.
```
hjoliver, line 58: This isn't very clear unless you carefully read the entire section, better to make it stand alone.
```suggestion
However, prior to 8.6 family names and globs would only match :term:`active tasks <active task>`.
```

src/reference/changes.rst
```diff
@@ -36,6 +36,95 @@ Cylc 8.5
...
+
+.. list-table::
+   :class: grid-table
+
+   * - **Cylc 7**
+     - **Cylc 8**
+   * - ::
+
+          cylc insert workflow FAMILY.cycle
+          cylc reset workflow FAMILY.cycle --state=waiting
```
hjoliver, line 100: In the interests of promoting how much better Cylc 8 is now, we should show what was needed to achieve the same thing *in general* in Cylc 7, not just in its still-in-the-task-pool sweet spot.  You've partially done that, with task insertion, but then we'd likely have to identify and trigger the first tasks that lead into the family, after setting them all up in the waiting state.

```suggestion
          cylc insert workflow FAMILY.cycle
          cylc reset workflow FAMILY.cycle --state=waiting
          # Trigger the task(s) that start the FAMILY sub-graph
          cylc trigger workflow member1.cycle 
```

hjoliver (approved):

wxtim (approved):

PR #883, opened 2025-11-05
Search error

wxtim:
Fix error where src/reference/job-script-vars/var-list.rst could be found in search results as <no-title>.

issue

Introduced by the renaming of `var-list.txt` to `var-list.rst`, causing the file to be indexed in https://github.com/cylc/cylc-doc/pull/853/files.

May require @jfrost-mo to confirm that this doesn't fundametally undo his work. 

Viewable at <redacted-internal-server>/doc/cylc/joe.mancell/html/search.html?q=CYLC_CYCLING_MODE&check_keywords=yes&area=default

**Requirements check-list**


jfrost-mo (changes requested):
The include doesn't seem to be working on `src/user-guide/writing-workflows/runtime.rst`. I do wonder if the fix is actually just to stop including the list in two placed, and just link to the reference page there instead.

That way this var-list.txt file can just actually be put into the `src/reference/job-script-vars/index.rst` file, keeping things simple.

src/reference/job-script-vars/index.rst
oliver-sanders: Wait, hang on.

Why do we have this random file floating in the reference section in the first place?!

Screenshot from 2025-11-13 12-44-43

Can't we just delete it?

wxtim: Off topic surely?

The blame says you're the only person in the northern hemisphere to edit that doc.

wxtim: Sorry, misunderstood your comment.. thought you were talking about 
image

I assume that job-script-environment variables is top level in the reference section because it's super useful...

IMO I'd quite like it if we had the top level looking like
- Reference: Workflow Config
- Reference: Task Job Environment Vars
- Reference: General

Since the first are 95% of the time what I'm looking for, and I'd expect that to rise to 99%+ for users. Either that or make those two sections the first items in the reference section with their own first level headings within it.

oliver-sanders (approved):

jfrost-mo (commented):
Clearing the review request.

PR #909, opened 2026-02-04
Various small proposed improvements

### LINKED ISSUE #924: The order of tutorials has gotten a bit confused.
The tutorials under the configuration-consolidation section were re-worked slightly but at the same time, the order was changed.  It seems like there is some confusion now as to the intended order.  As some of the tutorials continue work on the same file, this makes it hard to follow.

Currently, the order of the tutorials, including the precursors to configuration-consolidation, are listed as:

datetime-cycling (previous section) > runtime-configuration (previous section) > JinJa2 > Families > Parameters

However, at the top of both the Families and Parameters tutorials, it states:

> This practical continues on from the jinja2 practical

Additionally, during the JinJa2 tutorial, the flow.cylc file being worked on is altered to replace `resolution = 0.2` with `resolution = {{ RESOLUTION }}` with a JinJa2 variable set at the start.  This change is intended to continue onto the next tutorial for Families, and the help text for the tutorial suggests as much.  However, running `cylc get-resources tutorial/consolidation-tutorial` currently serves a previous version of the tutorial which does not include the JinJa2 resolution variable change.

It seems @MetRonnie corrected this [here]( https://github.com/cylc/cylc-flow/pull/7238), which adds the JinJa2 resolution variable to the Families tutorial, however, that change is not yet live and it still leaves a question as to the final ordering.

Discussing with @oliver-sanders, he thinks the order should be reverted to:
datetime-cycling (previous section) > runtime-configuration (previous section) > Families > JinJa2 > Parameters

and then fix the 
> This practical continues on from the xyz...

lines to be accurate. 

I believe to achieve this I need to:

1. Re-order the toctree on cylc-doc in all places to: Families > JinJa2 > Parameters (will be sorted in #909)
2. Update the Tutorial text on cylc-doc to not show JinJa2 being used for resolution in the Families tutorial (will be sorted in #909)
3. Update [this file](https://github.com/cylc/cylc-flow/blob/master/cylc/flow/etc/tutorial/consolidation-tutorial/flow.cylc) reverting the PR [here]( https://github.com/cylc/cylc-flow/pull/7238) so resolution is no longer set by JinJa2
4. Re-do the JinJa2 tutorial so it makes sense following the Families tutorial
5. Re-do the Parameters tutorial so it makes sense following the changes to Families and JinJa2 tutorials

samuel-denton:
https://wwwspice/~samuel.denton/cylc-doc/batch2_thoughts_on_docs/html/index.html

**Requirements check-list**

- Fixed a mistake on the interventions page causing tabs not to sync properly.
- Adding tabs to the intro page to show different UI types.  I find the current image a bit intimidating.  
- Added a note to Basic Cycling - Repeating Workflows to describe the reoccurrence in the example.


src/user-guide/interventions/index.rst
```diff
@@ -113,7 +113,7 @@ Re-Run Multiple Tasks
          :width: 75%
 
    .. tab-item:: CLI
-      :sync: gui
+      :sync: cli
```
samuel-denton, line 116: Small mistake here caused tabs to sync incorrectly.

src/tutorial/scheduling/integer-cycling.rst
```diff
@@ -71,7 +71,16 @@ To make a workflow repeat we must tell Cylc three things:
...
-     repeats at every integer :term:`cycle point`.
+     tells Cylc to repeat at every integer :term:`cycle point`.
+
+   .. note::
+
+      The recurrence term tells Cylc *how often* to repeat the tasks and 
+      dependencies in that block.  ``R1`` means *run once* (no cycling), 
+      ``P1`` means *repeat every cycle*, ``P2`` means *repeat every other cycle*, 
+      and so on.
+      Other terms are configurable and are introduced in the next section.
```
samuel-denton: I would like to add a note here explaining why R1 was swapped with P1 and illuding to more complex recurrences in further tutorials.
I am not yet happy with this wording however so will have another think.

src/tutorial/scheduling/integer-cycling.rst
```diff
@@ -71,7 +71,16 @@ To make a workflow repeat we must tell Cylc three things:
    * ``initial cycle point = 1`` tells Cylc to start counting cycle points
      from 1.
    * ``P1`` is the :term:`recurrence`; a ``P1`` :term:`graph string`
-     repeats at every integer :term:`cycle point`.
+     tells Cylc to repeat at every integer :term:`cycle point`.
```
samuel-denton: Minor change to keep consistency with other bullets.

src/installation.rst
```diff
@@ -19,19 +19,6 @@ Cylc runs on Unix-like systems including Linux and Mac OS.
...
-
-   We recommend using Mamba to install Cylc.
-   Mamba (or Micromamba) can be used as
-   `a faster, drop-in replacement for the conda command
-   <https://mamba.readthedocs.io/en/latest/index.html>`_.
-   If using Conda, make sure to use the libmamba-solver by updating to
-   Conda 23.10 or using the `conda-libmamba-solver plugin for conda
-   <https://conda.github.io/conda-libmamba-solver/getting-started/>`_.
-   The classic conda environment solver may be too slow for a complex package
-   like Cylc.
```
samuel-denton, line 32: https://github.com/cylc/cylc-doc/issues/916#issuecomment-4074141585

This tip is no longer relevant thanks to conda adopting the Libmamba dependency resolver

src/tutorial/furthertopics/clock-triggered-tasks.rst
```diff
@@ -60,7 +60,7 @@ Paste the following code into a ``flow.cylc`` file:
            [[[events]]]
                mail events = failed
        [[bell]]
-           script = printf 'bong%.02d\n' $(seq 1 $(cylc cyclepoint --print-hour))
+           script = printf 'bong%.02d\n' $(seq 1 $(cylc cyclepoint --print-hour)); sleep 5
```
samuel-denton, line 63: Added sleep 5 to tutorial to allow users time to see the workflow running.  A previous update to cylc sped up the running of this workflow significantly.

src/tutorial/furthertopics/clock-triggered-tasks.rst
```diff
@@ -77,7 +77,11 @@ Run your workflow using::
 Stop the workflow after a few cycles using ``cylc stop --now --now clock-trigger``.
 Notice how the tasks run as soon as possible rather than
 waiting for the actual time to be equal to the cycle point.
-
+If we didn't have the ``sleep 5`` in the task script,
+the tasks would run so quickly that we would not be able to see this behaviour.
+In some scenarios, such as climate modelling, this is the desired behaviour as
+it allows the workflow to run faster than real-time, what you might call simulation time.
+However, below we describe how to make the workflow run in real-time using clock triggers.
```
samuel-denton: Its a bit wordy at the moment, but a quick explainer as to when you might want cylc to trigger as many cylcs at once, for example to simulate events faster than real time for climate modelling.

src/tutorial/runtime/configuration-consolidation/index.rst
```diff
@@ -154,12 +156,12 @@ Which Approach To Use
...
-     particular machine or groups of tasks belonging to a particular system.
    * `Jinja2`_ is good at configuring settings which apply to the entire workflow
      rather than just a single task, as we can define variables then use them
      throughout the workflow.
+   * :term:`Families <family>` work best consolidating runtime configuration by
+     collecting tasks into broad groups, e.g. groups of tasks which run on a
+     particular machine or groups of tasks belonging to a particular system.
    * :term:`Parameterization <parameterization>` works best for describing tasks
      which are very similar but which have subtly different configurations
      (e.g. different arguments or environment variables).
```
samuel-denton, line 167: Reorder to match toc tree

src/tutorial/introduction.rst
```diff
@@ -88,10 +88,29 @@ Cylc can automatically:
...
+
+      .. image:: ../../src/img/tui-1.png
+         :alt: A screenshot of the Cylc TUI.
+         :width: 75%
+
+   .. tab-item:: CLI
+
+      .. image:: ../../src/img/cylc-cli.png
+         :alt: A screenshot of the Cylc CLI.
+         :width: 75%
```
samuel-denton: I think these tabs are a little clearer and less intimidating than the combined image used before.

src/installation.rst
samuel-denton, line 1: Lots of small wording changes to clarify what is and is not installed via pip vs conda.

ChrisPaulBennett (approved):
LGTM

src/tutorial/furthertopics/clock-triggered-tasks.rst
oliver-sanders: This makes it sound like the sleep is functional in the example, which it isn't.

Happy with the `sleep 5`, but not sure we need to provide an explanation. Happy to leave this implicit.

src/tutorial/runtime/configuration-consolidation/index.rst
```diff
@@ -139,10 +139,12 @@ The Three Approaches
...
 
-* :ref:`families <tutorial-cylc-families>`
-* :ref:`jinja2 <tutorial-cylc-jinja2>`
-* :ref:`parameters <tutorial-cylc-parameterization>`
+.. toctree::
+   :maxdepth: 1
 
+   jinja2
+   families
+   parameters
```
oliver-sanders: This changes the order in which the tutorials are attempted.

Does this work?

src/tutorial/scheduling/integer-cycling.rst
```diff
@@ -71,7 +71,16 @@ To make a workflow repeat we must tell Cylc three things:
...
+
+   .. note::
+
+      The recurrence term tells Cylc *how often* to repeat the tasks and 
+      dependencies in that block.  ``R1`` means *run once* (no cycling), 
+      ``P1`` means *repeat every cycle*, ``P2`` means *repeat every other cycle*, 
+      and so on.
+      Other terms are configurable and are introduced in the next section.
+
```
oliver-sanders: We explain the `P<x>` syntax in the next section, so we only need to explain `R1` here and describe what `P1` does.

Could reduce to something along the lines of this?:

```suggestion
     tells Cylc to repeat at every integer :term:`cycle point`
     (whereas ``R1`` means "run once").
```

src/tutorial/furthertopics/clock-triggered-tasks.rst
samuel-denton:
```suggestion
```

src/tutorial/runtime/configuration-consolidation/index.rst
samuel-denton: I believe this is just correcting the order in this sub-list so it matches the order elsewhere.  This is section 3 in the image:

Image

src/tutorial/scheduling/integer-cycling.rst
samuel-denton: This has been simplified as much was present in the next section.

src/tutorial/furthertopics/clock-triggered-tasks.rst
samuel-denton: Removed the explanation.

samuel-denton:
Checks failing due to rate limit during linkcheck.

src/tutorial/runtime/configuration-consolidation/index.rst
oliver-sanders: Try attempting the Jinja2 tutorial without first attempting the Families tutorial. There's a dependency between them.

The step numbers in the practicals outline the intended order:
* Families 1 & 2.
* Jinja2 3.
* Parameters 4.

The previous order was correct, i.e, it's the bit above which needs to change for consistency.

oliver-sanders:
> Checks failing due to rate limit during linkcheck.

The linkcheck test did fail, but not due to rate limiting. The message was:

```
Failed to establish a new connection: [Errno 101] Network is unreachable"))
```

I.e, the site was down.

src/tutorial/runtime/configuration-consolidation/index.rst
samuel-denton: Ah ok, I'll have another look.
If that is the case, and I think you are right, we are going to need to alter the toctree in a few places as currently, both the contents page and the menu go JinJa2 > Fam > Params.  Currently, going page by page from the start will be JinJa2 > Fam.
Let me confirm the order 100% and I'll make that change.

samuel-denton:
> > Checks failing due to rate limit during linkcheck.
> 
> The linkcheck test did fail, but not due to rate limiting. The message was:
> 
> ```
> Failed to establish a new connection: [Errno 101] Network is unreachable"))
> ```
> 
> I.e, the site was down.

Yep you're correct, the error starts: "Max retries exceeded with url" which I misinterpreted as a rate limit

Do we need to investigate / correct this URL?

samuel-denton:
Getting there:
https://wwwspice/~samuel.denton/cylc-doc/batch2_thoughts_on_docs/html/tutorial/runtime/configuration-consolidation/index.html

Still need to run through the tutorials again to check this all makes sense.  I've definitely missed some bits.

src/tutorial/runtime/configuration-consolidation/parameters.rst
```diff
@@ -182,7 +182,7 @@ Parameters can be either strings or integers:
 .. ifslides::
 
    .. rubric:: This practical continues on from the
-      :ref:`Families practical <cylc-tutorial-families-practical>`.
+      :ref:`Jinja2 practical <cylc-tutorial-jinja2-practical>`.
 
    Next section: :ref:`Which approach to use
    <cylc-tutorial-consolidation-conclusion>`
```
samuel-denton, line 188: It's currently inconsistent which pages make use of the ifslides / ifnotslides blocks.  Are these still used for anything?  I understood they were previously used for allowing the docs to generate in other formats but not sure if anything still uses them @oliver-sanders ?  I can make an effort to clean them up if not, or add any missing if they are need.  The JinJa2 page for example is missing some compared to the parametrization or families pages.

src/tutorial/runtime/configuration-consolidation/index.rst
samuel-denton: GH should be showing this as outdated as the change is reverted in a following commit.

samuel-denton:
I have run through the tutorials multiple times and I think I am now happy with the order and the accompanying text.  
Only thing I have a question about is the slightly inconsistent use of `ifslides / ifnotslides` but not sure thats relavent to this PR anyway.
Please review alongside https://github.com/cylc/cylc-flow/pull/7289 and close #924 if both are ok.

src/tutorial/runtime/configuration-consolidation/jinja2.rst
oliver-sanders: This is no longer true.

oliver-sanders: There is already a Jinja2 shebang line at the top of the file.

oliver-sanders (commented):
Found a couple of problems, having a think about how to make the Jinja2 part of the tutorial work...

Context:
* We used to have a `RESOLUTION` and an `API_KEY` variable.
* The `RESOLUTION` variable was de-duplicated in the Families tutorial.
* The `API_KEY` was de-duplicated in the Jinja2 tutorial.
* We moved away from Datapoint which removed the `API_KEY` variable.
* Tim changed the Jinja2 tutorial to de-duplicate the `RESOLUTION` variable, but this caused a conflict with the Family tutorial.

Unfortunately, post-reversion, we're left with the original problem, that the Jinaj2 tutorial no longer makes sense. Having a think about what we could do here without conflicting with the other tutorial parts...

src/tutorial/runtime/configuration-consolidation/jinja2.rst
samuel-denton: Good spot, I've added a commit to the cylc-flow PR removing the shebang:
https://github.com/cylc/cylc-flow/pull/7289/changes/46f249d851a829ee7e61decba0b18aca67b637fe
https://github.com/cylc/cylc-flow/pull/7289

samuel-denton: Ah yeah I did mean to update that text.  I was still deciding if I should just update the text to:

>       The ``RESOLUTION`` environment variable is used in this workflow.
>       Using Jinja2, we can create and set a value for a Jinja2 variable,  which could be re-used,
>       multiple times.  The main advantage over using other consolidation methods like families,
>       is that Jinja2 variables can be used anywhere in the workflow, not just the task definitions.
>       We could, for example, embed the RESOLUTION variable within a filename or task name.

Or update the tutorial to make use of the RESOLUTION variable multiple times somehow.
Which do you think?

samuel-denton: Having removed the jinja2 shebang, it's complaining that it's not there.  Do I need to add an exception for it somewhere?
https://github.com/cylc/cylc-flow/actions/runs/25436017737/job/74614084725#step:6:24

oliver-sanders: The shebang is in the file for another reason (see the `{% include` on the last line).

oliver-sanders: Not sure.

It worked nicely before as we had two variables to de-duplicate, one for each tutorial.

Having a look at other things we can Jinja'ise...

oliver-sanders: One idea I thought of:

```diff
diff --git a/cylc/flow/etc/tutorial/consolidation-tutorial/flow.cylc b/cylc/flow/etc/tutorial/consolidation-tutorial/flow.cylc
index eacfc98a8..b861cb8a9 100644
--- a/cylc/flow/etc/tutorial/consolidation-tutorial/flow.cylc
+++ b/cylc/flow/etc/tutorial/consolidation-tutorial/flow.cylc
@@ -2,6 +2,8 @@
 
 {% set RESOLUTION = 0.2 %}
 
+{% set WIND_CYCLES = [0, -3, -6] %}
+
 [scheduling]
     # Start the workflow 7 hours before now ignoring minutes and seconds
     # * previous(T-00) takes the current time ignoring minutes and seconds.
@@ -21,9 +23,9 @@
         # Repeat every six hours starting six hours after the initial
         # cycle point.
         +PT6H/PT6H = """
-            consolidate_observations => forecast
-            consolidate_observations[-PT3H] => forecast
-            consolidate_observations[-PT6H] => forecast
+{% for cycle in WIND_CYCLES %}
+            consolidate_observations[-PT{{ cycle }}H] => forecast
+{% endfor %}
             get_rainfall => forecast => post_process_exeter
         """
 
@@ -33,6 +35,13 @@

@@ -78,7 +83,7 @@
             # get substituted in the forecast script).
             WIND_FILE_TEMPLATE = $CYLC_WORKFLOW_WORK_DIR/{cycle}/consolidate_observations/wind_{xy}.csv
             # Offset in hours to cycles to process wind data from.
-            WIND_CYCLES = 0, -3, -6
+            WIND_CYCLES = {{ WIND_CYCLES | join(",") }}
             # The path to the rainfall file.
             RAINFALL_FILE = $CYLC_WORKFLOW_WORK_DIR/$CYLC_TASK_CYCLE_POINT/get_rainfall/rainfall.csv
             # Create the html map file in the task's log directory.
@@ -86,13 +91,5 @@

diff --git a/cylc/flow/etc/tutorial/cylc-forecasting-workflow/bin/forecast b/cylc/flow/etc/tutorial/cylc-forecasting-workflow/bin/forecast
index c5e46cbf8..32351b6ab 100755
--- a/cylc/flow/etc/tutorial/cylc-forecasting-workflow/bin/forecast
+++ b/cylc/flow/etc/tutorial/cylc-forecasting-workflow/bin/forecast
@@ -67,6 +67,7 @@ def get_wind_data(wind_file_path, wind_cycles):
     """
     wind_fields = []
     for cycle in wind_cycles:
+        cycle = int(cycle) * -1
         wind_x = util.read_csv(wind_file_path.format(cycle=cycle, xy='x'))
         wind_y = util.read_csv(wind_file_path.format(cycle=cycle, xy='y'))
         wind_fields.append((wind_x, wind_y))
```

Not sure I like it though.

samuel-denton: Ah yeah... I did look for Jinja2 in the file but ignored the includes line...
I've added the shebang back and will change the text instead.

samuel-denton: It's a good example, just not sure if it introduces too many new things at once.  
As a simple way to re-use the RESOLUTION variable could we modify a file name to include it?

samuel-denton, line 1: I have updated the jinja2 tutorial again to make more use of the resolution variable, is that sufficient @oliver-sanders?
Or would you prefer to make more use of jinja2 features introduced further up the page like loops?

ChrisPaulBennett (approved):

src/tutorial/introduction.rst
```diff
@@ -88,10 +88,29 @@ Cylc can automatically:
 
 Cylc provides a variety of command line and GUI tools for visualising,
 monitoring, and controlling workflows. The Cylc TUI (Terminal
-User Interface), web GUI, and ``cylc scan`` (bottom left) are shown below.
+User Interface), web GUI, and ``cylc scan`` are shown below.
 
-.. image:: /tutorial/img/cylc-tools.png
-   :alt: A screenshot of several Cylc tools.
+**Placeholder Images Currently Used**
+
```
oliver-sanders: I'm not sure what this title means, but I think the meaning is self-explanatory:

```suggestion
```

samuel-denton: Yep sorry, thats a note I left for myself to not mark the PR as ready until I have updated the images, which I have done I just forgot to remove the note...

src/tutorial/runtime/configuration-consolidation/jinja2.rst
oliver-sanders: Using the `RESOLUTION` variable to template a filename is a reasonable idea, but the user might be thinking, why are we including the resolution in the filename since we aren't using this for anything?

Also, later on, we get the user to open this file in Cylc Review, and in the Rose Tutorial, we get them to extract the map file into a Rose configuration.

Having had a think about it, here's the best thing I've come up with so far. It does some light maths, which is a legit use case for Jinja2 and something we genuinely couldn't handle with families/parameters.

```diff
diff --git a/cylc/flow/etc/tutorial/cylc-forecasting-workflow/flow.cylc b/cylc/flow/etc/tutorial/cylc-forecasting-workflow/flow.cylc
index 86437e6c0..aafc1358a 100644
--- a/cylc/flow/etc/tutorial/cylc-forecasting-workflow/flow.cylc
+++ b/cylc/flow/etc/tutorial/cylc-forecasting-workflow/flow.cylc
@@ -1,4 +1,9 @@
 #!jinja2
+
+{# Generate 5 forecasts at 60 minute intervals. #}
+{% FORECAST_LENGTH = 60 %}
+{% FORECAST_COUNT = 5 %}
+
 [task parameters]
     # A list of the weather stations we will be fetching observations from.
     station = camborne, heathrow, shetland, aldergrove
@@ -67,7 +72,7 @@
         script = get-rainfall
 
     [[forecast]]
-        script = forecast 60 5  # Generate 5 forecasts at 60 minute intervals.
+        script = forecast {{ FORECAST_LENGTH }} {{ FORECAST_COUNT }}
         [[[environment]]]
             # The path to the files containing wind data (the {variables} will
             # get substituted in the forecast script).
@@ -82,7 +87,8 @@
             MAP_TEMPLATE = "$CYLC_WORKFLOW_RUN_DIR/lib/template/map.html"
 
     [[post_process<site>]]
-        # Generate a forecast for the location <site> 60 minutes in the future.
-        script = post-process $CYLC_TASK_PARAM_site 60
+        # Generate a forecast for the location <site>
+        # {{ FORECAST_LENGTH * FORECAST_COUNT }} minutes in the future
+        script = post-process $CYLC_TASK_PARAM_site {{ FORECAST_LENGTH * FORECAST_COUNT }}
 
 {% include 'etc/python-job.settings' %}
```

The downside is that we would need to make some changes to the Rose tutorial to accommodate for this:

https://metomi.github.io/rose/doc/html/tutorial/rose/applications.html#admonition-0

samuel-denton: Yeah I like that. I was trying to find a change that wouldn't mean altering another downstream tutorial but it seems like that's inevitable.  

That example looks like its been implemented after the Parameterisation tutorial changes are made so need to make sure we have the order correct.  ```[[post_process<site>]]``` should still be ```[[post_process_exeter]]``` I believe.

oliver-sanders: Yes, this diff is from the finished tutorial (etc/tutorial/cylc-forecasting-workflow)

> [[post_process<site>]] should still be [[post_process_exeter]] I believe.

In the context of the Jinja2 tutorial, yes.

oliver-sanders: (FYI: I didn't put much thought into those `FORECAST_*` variable names)

samuel-denton: I need to go through from start to finish again but at the moment, I don't think any of the changes here will effect the Rose tutorials.  The one at:
https://metomi.github.io/rose/doc/html/tutorial/rose/applications.html#admonition-0
Is a standalone app, and the ones on:
https://wwwspice/~samuel.denton/rose-doc/consolidation_tutorial_update/html/tutorial/rose/suites.html
Effect parts of the workflow we haven't changed.
I will confirm this before marking the PR's as ready for review.

src/tutorial/runtime/configuration-consolidation/families.rst
```diff
@@ -310,8 +310,8 @@ The ``root`` Family
                   MAP_TEMPLATE = "$CYLC_WORKFLOW_RUN_DIR/lib/template/map.html"
 
           [[post_process_exeter]]
-              # Generate a forecast for Exeter 60 minutes into the future.
-              script = post-process exeter 60
+              # Generate a forecast for Exeter 300 minutes into the future.
+              script = post-process exeter 300
```
samuel-denton, line 273: This is to match the change in the Jinja2 tutorial.

src/tutorial/runtime/configuration-consolidation/jinja2.rst
samuel-denton: This suggestion has been implemented.

oliver-sanders:
There's a bit of a problem in tutorial/runtime/configuration-consolidation/families.

It starts like so:

> [Families](file:///home/users/oliver.sanders/cylc-doc/doc/8.6.4.dev/html/glossary.html#term-family) are groups of tasks which share a common configuration. In the present example the common configuration is:
>
> ```cylc
> script = get-observations
> [[[environment]]]
>     GET_NEARBY = true
> ```

However, there is no `GET_NEARBY` variable in the snippet on the previous change, or the previous tutorial.

The variables it then gets us to consolidate are `RESOLUTION` and `DOMAIN`:

> ```diff
> + [runtime]
> +    [[root]]
> +        [[[environment]]]
> +            # The dimensions of each grid cell in degrees.
> +            RESOLUTION = 0.2
> +            # The area to generate forecasts for (lng1, lat1, lng2, lat2).
> +            DOMAIN = -12,46,12,61  # Do not change!
> ```

samuel-denton:
> However, there is no `GET_NEARBY` variable in the snippet on the previous change, or the previous tutorial.

I had spotted that, but I though the explainer text was somewhat separate from the practicals.  The Jinja2 one for example talks about lots of things that are not touched in the practical.  However I can see that it could be confusing if someone was going page by page reading the whole thing as it uses the same examples as the practical.  I can look at re-wording that bit to describe one of the variables actually consolidated.

src/tutorial/runtime/configuration-consolidation/jinja2.rst
```diff
@@ -165,57 +165,57 @@ This would result in:
...
+        [[post_process_exeter]]
+        -   # Generate a forecast for Exeter 300 minutes in the future.
+        -   script = post-process exeter 300
+        +   # Generate a forecast {{ FORECAST_LENGTH * FORECAST_COUNT }} minutes in the future.
+        +   script = post-process exeter {{ FORECAST_LENGTH * FORECAST_COUNT }}
 
       Check the result with ``cylc config``. The Jinja2 will be processed
-      so you should not see any difference after making these changes.
+      so you should not see any difference after making these changes, other than
+      the MAP_FILE filename now including the resolution.
```
MetRonnie: I think the filename change is no longer present in the PR?

src/tutorial/scheduling/integer-cycling.rst
```diff
@@ -71,7 +71,8 @@ To make a workflow repeat we must tell Cylc three things:
    * ``initial cycle point = 1`` tells Cylc to start counting cycle points
      from 1.
    * ``P1`` is the :term:`recurrence`; a ``P1`` :term:`graph string`
-     repeats at every integer :term:`cycle point`.
+     tells Cylc to repeat at every integer :term:`cycle point`
```
MetRonnie:
```suggestion
     that tells Cylc to repeat at every integer :term:`cycle point`
```

src/tutorial/introduction.rst
MetRonnie: `cylc-tools.png` can be deleted as it is no longer used.

Also, can you remove the `width` directive as it makes the images smaller and harder to see properly?

The text above mentions `cylc scan` but it is not present in the CLI screenshot. Also the CLI screenshot text is rather small even with the `width` directive removed.

Sorry that was a bunch of alsos 😄

src/tutorial/runtime/configuration-consolidation/jinja2.rst
```diff
@@ -165,57 +165,57 @@ This would result in:
...
+      In contrast, Jinja2 can be used in any part of a workflow, and can be used
+      for simple programmatic logic.  Here, we will make use of Jinja2 for some
+      maths and string manipulation.
 
-      At the top of the :cylc:conf:`flow.cylc` file add the Jinja2 shebang line. Then
-      copy the value of the ``RESOLUTION`` environment variable and use it to
-      define an ``RESOLUTION`` Jinja2 variable:
+      At the top of the :cylc:conf:`flow.cylc` file you should see the Jinja2
+      shebang line has been included for you.  Create some new Jinja2 variables
+      for ``FORECAST_LENGTH`` and ``FORECAST_COUNT```:
```
MetRonnie:
```suggestion
      for ``FORECAST_LENGTH`` and ``FORECAST_COUNT``:
```

src/tutorial/scheduling/integer-cycling.rst
samuel-denton: Not sure I agree with that change:
- a P1 graph string tells cylc to ...
vs
- a P1 graph string that tells cylc to...

src/tutorial/runtime/configuration-consolidation/jinja2.rst
samuel-denton: Yep... I've changed this text so many times now I'm loosing track!

samuel-denton:
@oliver-sanders @MetRonnie 
I believe all comments have been addressed so should be ready for another look.

src/tutorial/runtime/configuration-consolidation/jinja2.rst
```diff
@@ -165,57 +165,56 @@ This would result in:
...
-         +          RESOLUTION = {{ RESOLUTION }}
-
-         [[post_process_exeter]]
-            # Generate a forecast for Exeter 60 minutes in the future.
-            script = post-process exeter 60
-            [[[environment]]]
-         -          RESOLUTION = 0.2
-         +          RESOLUTION = {{ RESOLUTION }}
+        [[forecast]]
+        -   script = forecast 60 5
```
oliver-sanders: Slight difference in the diff from my runthrough:

```suggestion
        -   script = forecast 60 5  # Generate 5 forecasts at 60 minute intervals.
```

src/tutorial/runtime/configuration-consolidation/parameters.rst
```diff
@@ -330,6 +330,7 @@ Parameters can be either strings or integers:
 
             -[[post_process_exeter]]
             +[[post_process<site>]]
-                 # Generate a forecast for Exeter 60 minutes in the future.
-            -    script = post-process exeter 60
-            +    script = post-process $CYLC_TASK_PARAM_site 60
+            -    # Generate a forecast for Exeter 300 minutes in the future.
+            -    script = post-process exeter 300
+            +    #  Generate a forecast {{ FORECAST_LENGTH * FORECAST_COUNT }} minutes in the future for <site>
+            +    script = post-process $CYLC_TASK_PARAM_site {{ FORECAST_LENGTH * FORECAST_COUNT }}
```
oliver-sanders: Slight difference in the diff from my runthrough:

```suggestion
                 # Generate a forecast {{ FORECAST_LENGTH * FORECAST_COUNT }} minutes in the future.
            -    script = post-process exeter {{ FORECAST_LENGTH * FORECAST_COUNT }}
            +    script = post-process "$CYLC_TASK_PARAM_site" {{ FORECAST_LENGTH * FORECAST_COUNT }}
```

src/tutorial/runtime/configuration-consolidation/families.rst
```diff
@@ -51,65 +49,69 @@ Runtime
...
+               RAINFALL_FILE = $CYLC_WORKFLOW_WORK_DIR/$CYLC_TASK_CYCLE_POINT/get_rainfall/rainfall.csv
+               MAP_FILE = "${CYLC_TASK_LOG_ROOT}-map.html"
+               MAP_TEMPLATE = "$CYLC_WORKFLOW_RUN_DIR/lib/template/map.html"
+   -           RESOLUTION = {{ RESOLUTION }}
+   -           DOMAIN = -12,46,12,61
+       [[post_process_exeter]]
+           script = post-process exeter 300
+   -       [[[environment]]]
+   -           RESOLUTION = {{ RESOLUTION }}
+   -           DOMAIN = -12,46,12,61
```
oliver-sanders: The diff is slightly misaligned.

But also a tad confusing given that we get them to do something different in the exercise.

I think the section has made the point about consolidation and this can be removed:

```suggestion
```

src/tutorial/scheduling/integer-cycling.rst
```diff
@@ -71,7 +71,8 @@ To make a workflow repeat we must tell Cylc three things:
...
+     (whereas ``R1`` means "run once").
```
MetRonnie: Instead of https://github.com/cylc/cylc-doc/pull/909#discussion_r3287967181, trying to improve the readability of this somewhat
```rst
   * ``P1`` is the :term:`recurrence`. The tasks in a ``P1``
     :term:`graph string` repeat at every integer :term:`cycle point`
     (whereas ``R1`` means "run once").

```

src/tutorial/runtime/configuration-consolidation/families.rst
samuel-denton: Example removed

src/tutorial/scheduling/integer-cycling.rst
samuel-denton: Changed the wording for clarity

MetRonnie (approved):
Thanks, this looks better now 👍 I don't have much time to go through the tutorials to check the code diffs etc are completely consistent with what you get, if someone could do so that would be great

oliver-sanders:
(test failure unrelated)

oliver-sanders (approved):
:+1:

oliver-sanders:
(kicking CI)

PR #911, opened 2026-02-13
troubleshooting: document failed/TERM

oliver-sanders:
**Requirements check-list**


wxtim (approved):
Useful clarification.

What would you expect to see if the Job runner has become uncontactable, but the scheduler knows it has exceeded ETL?

oliver-sanders:
> What would you expect to see if the Job runner has become uncontactable, but the scheduler knows it has exceeded ETL?

Scheduler will issue a poll and retrieve the failed output.

The batch system will *likely* kill the job with XCPU, but might not.

PR #920, opened 2026-04-13
Force external links to open in a new tab

### LINKED ISSUE #908: External links opening in new tabs - Sphinx/reST constraints and options
In the Cylc documentation, both internal and external links open in the same browser tab.

After looking into what’s possible with reStructuredText and Sphinx, it appears to be a deliberate design choice that Sphinx does not provide a native way to control whether links open in a new tab/window.

From reviewing GitHub issues and community discussions, the main reasons given for not supporting this are:

- Opening new tabs/windows is considered a user agent decision
- Forcing new tabs can have accessibility implications
- Documentation is intended to be medium-agnostic (HTML, PDF, man pages, etc.)
- Users can choose to open links in new tabs themselves
- JavaScript-based workarounds are considered sufficient where required

The main discussions can be found here:

- https://github.com/sphinx-doc/sphinx/issues/12477
- https://github.com/sphinx-doc/sphinx/issues/1634

Many of these arguments could reasonably be applied to our own documentation, so whether external links should open in a new tab by default is a policy decision for discussion.

If we did want this behaviour, there are a few known workarounds:

1. **JavaScript workaround** (recommended by Sphinx maintainers)
Sphinx already marks external links with the external CSS class, so JavaScript can be used to target only those links and force them to open in a new tab. This would be a global behaviour change for all external links.  This solution can be implemented in a few ways, as shown here: [open-a-link-in-a-new-window-in-restructuredtext](https://stackoverflow.com/questions/11716781/open-a-link-in-a-new-window-in-restructuredtext)

    - Vanilla JS Solution
    ```j
        $(document).ready(function () {
           $('a[href^="http://"], a[href^="https://"]').not('a[class*=internal]').attr('target', '_blank');
        }); 
    ```

    - JQuery solution:

    ```j
        $(document).ready(function () {
            $('a.external').attr('target', '_blank');
        });
    ```

2. **HTML workaround**
RST allows inserts of HTML blocks, so those links could be replaced with `<a>` tags including the parameter `target="_blank"`.
    - This doesn't seem to be a viable solution due to possible issues parsing the documentation into other formats.
3. **Sphinx plugins**
There is a plugin for Sphinx that just adds this functionality: [ftnext/sphinx-new-tab-link](https://github.com/ftnext/sphinx-new-tab-link) which seems to be maintained and fairly well used.
    - It appears to be maintained and reasonably well-used, but I’m not currently sure whether adding third-party Sphinx extensions aligns with our guidelines / practices.

So it seems we can implement this change if it's decided to be appropriate.

samuel-denton:
Closes #908 if agreed.

Added some JavaScript to the theme layout which Adds target=_blank to external links so they open in a new tab.
I have seen some discourse online as to whether developers should try to effect the users choice when clicking links (they can already chose to open links in a new tab themselves), however in our case, external links are almost always definitions or further reading type of stuff so it feels a bit jarring for them to take you away from the docs.

Also added an icon and hover text to external links making them obvious.

Built example:
https://wwwspice/~samuel.denton/cylc-doc/8.6.x_Ext_links_new_tab/html/tutorial/scheduling/graphing.html
image

**Requirements check-list**

### DIFF
--- a/src/_static/css/custom.css
+++ b/src/_static/css/custom.css
@@ -20,10 +20,12 @@
   height: 7em !important;
 }
 
-/* Add an icon to external links */
-a.external::after {
-  font-size: small;
+/* Add an icon to external links, excluding those with the "image-reference" class */
+a.external:not(.image-reference)::after {
+  font-size: 0.7em;
   font-weight: bolder;
-  content: "↗";
   margin-left: 0.3em;
+  vertical-align: super;
+  opacity: 0.5;
+  content: "↗";
 }
--- a/src/_static/css/custom.css
+++ b/src/_static/css/custom.css
@@ -22,7 +22,8 @@
 
 /* Add an icon to external links */
 a.external::after {
-  content: "🔗 ↗ ⧉";
+  font-size: small;
+  font-weight: bolder;
+  content: "↗";
   margin-left: 0.3em;
-  color: black;
 }
--- a/src/_static/js/extern_links_new_tab.js
+++ b/src/_static/js/extern_links_new_tab.js
@@ -25,7 +25,7 @@ document.addEventListener("DOMContentLoaded", function () {
     document.querySelectorAll('a.external').forEach(link => {
         link.setAttribute('target', '_blank');
         link.setAttribute('rel', 'noopener noreferrer');
-        link.setAttribute('title', 'This link will open in a new tab and take you to an external site.');
+        link.setAttribute('title', 'Opens in a new tab.');
         link.setAttribute('aria-label', 'External link: opens in a new tab');
     });
 });
--- a/src/_static/js/extern_links_new_tab.js
+++ b/src/_static/js/extern_links_new_tab.js
@@ -15,14 +15,14 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  * ------------------------------------------------------------------------- */
 
-/* Fix zero height auto size of svg */
-.logo {
-  height: 7em !important;
-}
-
-/* Add an icon to external links */
-a.external::after {
-  content: "🔗 ↗ ⧉";
-  margin-left: 0.3em;
-  color: black;
-}
+/*
+    Adds target=_blank to external links so they open in a new tab:
+    - noopener: prevents the new page accessing window.opener (tabnabbing protection)
+    - noreferrer: removes referrer info and adds extra safety
+*/
+document.addEventListener("DOMContentLoaded", function () {
+    document.querySelectorAll('a.external').forEach(link => {
+        link.setAttribute('target', '_blank');
+        link.setAttribute('rel', 'noopener noreferrer');
+    });
+});
--- a/src/_static/js/extern_links_new_tab.js
+++ b/src/_static/js/extern_links_new_tab.js
@@ -24,5 +24,7 @@ document.addEventListener("DOMContentLoaded", function () {
     document.querySelectorAll('a.external').forEach(link => {
         link.setAttribute('target', '_blank');
         link.setAttribute('rel', 'noopener noreferrer');
+        link.setAttribute('title', 'This link will open in a new tab and take you to an external site.');
+        link.setAttribute('aria-label', 'External link: opens in a new tab');
     });
 });
--- a/src/_static/js/extern_links_new_tab.js
+++ b/src/_static/js/extern_links_new_tab.js
@@ -19,6 +19,7 @@
     Adds target=_blank to external links so they open in a new tab:
     - noopener: prevents the new page accessing window.opener (tabnabbing protection)
     - noreferrer: removes referrer info and adds extra safety
+    - title and aria-label: provide context for screen readers and tooltips
 */
 document.addEventListener("DOMContentLoaded", function () {
     document.querySelectorAll('a.external').forEach(link => {
--- a/src/_templates/layout.html
+++ b/src/_templates/layout.html
@@ -1,1 +1,14 @@
+<script type="text/javascript">
+    /*
+        Adds target=_blank to external links so they open in a new tab:
+        - noopener: prevents the new page accessing window.opener (tabnabbing protection)
+        - noreferrer: removes referrer info and adds extra safety
+    */
+    document.addEventListener("DOMContentLoaded", function () {
+        document.querySelectorAll('a.external').forEach(link => {
+            link.setAttribute('target', '_blank');
+            link.setAttribute('rel', 'noopener noreferrer');
+        });
+    });
+</script>
 {% extends "!layout.html" %}
--- a/src/_templates/layout.html
+++ b/src/_templates/layout.html
@@ -1,14 +1,1 @@
-<script type="text/javascript">
-    /*
-        Adds target=_blank to external links so they open in a new tab:
-        - noopener: prevents the new page accessing window.opener (tabnabbing protection)
-        - noreferrer: removes referrer info and adds extra safety
-    */
-    document.addEventListener("DOMContentLoaded", function () {
-        document.querySelectorAll('a.external').forEach(link => {
-            link.setAttribute('target', '_blank');
-            link.setAttribute('rel', 'noopener noreferrer');
-        });
-    });
-</script>
 {% extends "!layout.html" %}
--- a/src/conf.py
+++ b/src/conf.py
@@ -239,18 +239,29 @@ html_context = {
     'sidebar_version_name': None,  # override name in version picker
 }
 
-html_static_path = ['_static']
-
-# These paths are either relative to html_static_path
-# or fully qualified paths (eg. https://...)
-html_css_files = ['css/custom.css']
-
 # Disable timestamp otherwise inserted at bottom of every page.
 html_last_updated_fmt = ''
 
 # Output file base name for HTML help builder.
 htmlhelp_basename = 'cylcdoc'
 
+# -- Static file locations -----------------------------------------------
+html_static_path = ['_static']
+
+# These paths are either relative to html_static_path
+# or fully qualified paths (eg. https://...)
+# NOTE: these are added to the end of the default/theme static files, so
+# will override them if there are any conflicts.
+
+# Custom CSS
+html_css_files = [
+    'css/custom.css',
+]
+
+# Custom JS
+html_js_files = [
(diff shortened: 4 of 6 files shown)


src/_templates/layout.html
```diff
@@ -1 +1,14 @@
...
+        Adds target=_blank to external links so they open in a new tab:
+        - noopener: prevents the new page accessing window.opener (tabnabbing protection)
+        - noreferrer: removes referrer info and adds extra safety
+    */
+    document.addEventListener("DOMContentLoaded", function () {
+        document.querySelectorAll('a.external').forEach(link => {
+            link.setAttribute('target', '_blank');
+            link.setAttribute('rel', 'noopener noreferrer');
+        });
+    });
```
samuel-denton: There are two places we can insert this script, either here in the theme layout doc or in a custom JS doc.  The advantage of putting it here is that this file already exists as part of the RTD theme.  The downside being it would potentially need re-implementing if the theme was changed.

Moving this to a custom JS file is as simple as adding a new file to src/_static/js with that function, then add:

```
html_js_files = [
    'name_of_new_file.js',
]
```

To src/conf.py

samuel-denton:
This is working [here](https://wwwspice/~samuel.denton/cylc-doc/8.6.x_Ext_links_new_tab/html/tutorial/scheduling/graphing.html)
image
Link 1 is internal, link 2 is external.

samuel-denton:
The other method I found to achieve the same was [this](https://github.com/ftnext/sphinx-new-tab-link) extension, which worked but relies on it being maintained.  The only additional benefit from the extension is that it can display an icon next to external links.  We could replicate that if we wanted.

samuel-denton:
If we want per link control, we could use something like this:
```rst
A Cylc workflow is defined by a flow.cylc configuration file, which uses a nested |location_link|.

.. |location_link| raw:: html

   <a href="https://en.wikipedia.org/wiki/INI_file" target="_blank">INI format</a>
```
But that's messy and would be a pain to implement for every link we want to effect.

oliver-sanders:
LGTM.

Adding the icon might be simple with a drop of CSS.

E.g, This will cause a red "x" symbol to be added after each external link:

```css
a.external:after {
  content: ' x';
  color:red;
}
```

I think the RTD theme imports "Font Awesome" which probably has an appropriate icon, so might be simple.

samuel-denton:
> Adding the icon might be simple with a drop of CSS.

@oliver-sanders  I can do that.  Do you think the JS should be in the theme layout.html file or should I move it to _static/js like the CSS?

oliver-sanders:
> Do you think the JS should be in the theme layout.html file or should I move it to _static/js like the CSS?

I'm not actually sure how the `_static` scripts get sourced, you may find you need to add a corresponding `<script source="..." />` line to the HTML file anyway?

samuel-denton:
> > Do you think the JS should be in the theme layout.html file or should I move it to _static/js like the CSS?
> 
> I'm not actually sure how the `_static` scripts get sourced, you may find you need to add a corresponding `<script source="..." />` line to the HTML file anyway?

@oliver-sanders  I believe I can add:
```
html_js_files = [
    'name_of_new_file.js',
]
```
to the conf.py to pick up scripts.  That seems like the proper way to do it and it opens the door for any further JS we might want to add, it's just whether it's worth creating a new file in this case.

oliver-sanders:
Sounds good.

New files cost little.

src/_static/css/custom.css
```diff
@@ -19,3 +19,10 @@
 .logo {
   height: 7em !important;
 }
+
+/* Add an icon to external links */
+a.external::after {
+  content: "🔗 ↗ ⧉";
```
samuel-denton: I've looked into using font awesome icons but it looks like on the latest versions of the RTD theme for sphinx, it only bundles specific bits of RTD, so I have been unable to get any icons from them working.  I would have liked: [this](https://fontawesome.com/v6/icons/arrow-up-right-from-square?f=classic&s=solid)
So assuming FA icons are not possible, which of those three do you like @oliver-sanders?  I will adjust the colour, size and position once we pick one.

src/_static/js/extern_links_new_tab.js
```diff
@@ -0,0 +1,28 @@
...
+    Adds target=_blank to external links so they open in a new tab:
+    - noopener: prevents the new page accessing window.opener (tabnabbing protection)
+    - noreferrer: removes referrer info and adds extra safety
+*/
+document.addEventListener("DOMContentLoaded", function () {
+    document.querySelectorAll('a.external').forEach(link => {
+        link.setAttribute('target', '_blank');
+        link.setAttribute('rel', 'noopener noreferrer');
+    });
+});
```
samuel-denton, line 31: Not sure on preferred code/comment style for JS yet, hopefully this is ok?

src/conf.py
```diff
@@ -239,18 +239,29 @@ def generate_task_icon_modifier_rst():
...
+
+# Custom CSS
+html_css_files = [
+    'css/custom.css',
+]
+
+# Custom JS
+html_js_files = [
+    'js/extern_links_new_tab.js',
+]
```
samuel-denton, line 264: Moved the static file settings to their own block.

src/_static/css/custom.css
samuel-denton: image

Options ^^

oliver-sanders: Fair 'nough.

Would probs go with ↗ 

We can mess with the `font-size` and `vertical-align` to make it display nicely.

src/_static/js/extern_links_new_tab.js
oliver-sanders: Fine by me.

We generally use eslint "standard" for JS.

oliver-sanders:
Got an example of unwanted decoration on the homepage:

Screenshot from 2026-04-15 11-10-46


And a suggested change to the CSS:

```css
font-size: 0.7em;
vertical-align: super;
opacity: 0.5;
```

Screenshot from 2026-04-15 11-12-01

oliver-sanders:
Curiously, I found some internal links flagged as external?

Screenshot from 2026-04-15 15-54-25

workflow-design-guide/style-guide.html#indentation

samuel-denton:
> Curiously, I found some internal links flagged as external?

ok interesting, I'll have a look into why they are tagged differently.  I think there is some config that controls what counts as external.

I'm thinking about adding some hover text that tells the user a link is going to take them to an external page too.

samuel-denton:
> Curiously, I found some internal links flagged as external?

@oliver-sanders 
How did you build / view the test site?  That section is added by:
```
.. automodule:: cylc.flow.scripts.lint
```
Which is syntax I am not yet familiar with but appears to either generate or import a module from elsewhere.  The links within this module **are external** when built / viewed on my test server as they link back to the live version of the site.

URL of the test site: 
https://wwwspice/~samuel.denton/cylc-doc/8.6.x_Ext_links_new_tab/html/user-guide/writing-workflows/configuration.html

URL that is linked to: 
https://cylc.github.io/cylc-doc/stable/html/workflow-design-guide/style-guide.html#tab-characters

So I would guess this will fix itself on a live build.  How we test that I am unsure.
If we wanted to explicitly fix this, we could exclude links where the href contains something like:
'cylc.github.io/cylc-doc/stable/html'

oliver-sanders:
> How did you build / view the test site

`make html && firefox ./build`

oliver-sanders:
`automodule` and friends are Sphinx's autodocumenter tools, they turn Python docstrings into Sphinx documentation.

In this case, it's reading from `cylc.flow.scripts.lint.__doc__` which, on inspection, reveals the issue.

Don't worry about this one, I'll poke at it from the cylc-flow end...

oliver-sanders:
> Don't worry about this one, I'll poke at it from the cylc-flow end...

No clean way to handle this one, will just have to ignore it.

src/_static/js/extern_links_new_tab.js
```diff
@@ -0,0 +1,30 @@
...
+/*
+    Adds target=_blank to external links so they open in a new tab:
+    - noopener: prevents the new page accessing window.opener (tabnabbing protection)
+    - noreferrer: removes referrer info and adds extra safety
+*/
+document.addEventListener("DOMContentLoaded", function () {
+    document.querySelectorAll('a.external').forEach(link => {
+        link.setAttribute('target', '_blank');
+        link.setAttribute('rel', 'noopener noreferrer');
+        link.setAttribute('title', 'This link will open in a new tab and take you to an external site.');
```
samuel-denton: adds on hover text.

src/_static/js/extern_links_new_tab.js
```diff
@@ -24,5 +24,7 @@ document.addEventListener("DOMContentLoaded", function () {
     document.querySelectorAll('a.external').forEach(link => {
         link.setAttribute('target', '_blank');
         link.setAttribute('rel', 'noopener noreferrer');
+        link.setAttribute('title', 'This link will open in a new tab and take you to an external site.');
```
oliver-sanders: If doing hover over, would probably suggest shorter text, e.g, `open in new tab` or `external link: open in new tab` or whatever.

src/_static/js/extern_links_new_tab.js
```diff
@@ -0,0 +1,30 @@
...
+    Adds target=_blank to external links so they open in a new tab:
+    - noopener: prevents the new page accessing window.opener (tabnabbing protection)
+    - noreferrer: removes referrer info and adds extra safety
+*/
+document.addEventListener("DOMContentLoaded", function () {
+    document.querySelectorAll('a.external').forEach(link => {
+        link.setAttribute('target', '_blank');
+        link.setAttribute('rel', 'noopener noreferrer');
+        link.setAttribute('title', 'This link will open in a new tab and take you to an external site.');
+        link.setAttribute('aria-label', 'External link: opens in a new tab');
```
samuel-denton, line 29: additional information for a screen reader.

src/_static/js/extern_links_new_tab.js
samuel-denton: Applied that suggestion

RyanBcoding (approved):
Tested changes. Seems to be working as expected. Comments explain decision for why the changes were made, so happy to approve.

oliver-sanders (approved):

PR #923, opened 2026-04-27
fix workflow

oliver-sanders:
3.x is not a valid Python version in Conda.

**Requirements check-list**

### DIFF
--- a/.github/workflows/create-conda-envs.yml
+++ b/.github/workflows/create-conda-envs.yml
@@ -25,7 +25,7 @@ jobs:
     timeout-minutes: 30
     strategy:
       matrix:
-        python: ['3.x']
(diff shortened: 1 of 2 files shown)


.github/workflows/create-conda-envs.yml
MetRonnie: Edit: code suggestion doesn't work properly on non-modified lines, but my suggestion is to remove these lines

MetRonnie (changes requested):

.github/workflows/create-conda-envs.yml
oliver-sanders: Yeah, that would probably help!

oliver-sanders:
And onto the next problem.

PR #929, opened 2026-05-21
Clarify install target docs.

hjoliver:
Companion of https://github.com/cylc/cylc-flow/pull/7315

Current docs are a bit misleading (IMO) on exactly what an `install target` is.

Question: can we also change `install target = localhost` in various example to something else, e.g. `local_cluster` to make it more obvious that an install target is not a hostname?

**Requirements check-list**

### DIFF
--- a/src/reference/config/writing-platform-configs.rst
+++ b/src/reference/config/writing-platform-configs.rst
@@ -41,7 +41,8 @@ Platforms define settings, most importantly:
 
  - A set of ``hosts``.
  - A ``job runner`` (formerly a ``batch system``) where Cylc can submit a job.
- - An ``install target`` for Cylc to install job files on.
+ - An ``install target``, representing each distinct filesystem that Cylc might
+   need to install workflow files on.
 
 Why Were Platforms Introduced?
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
@@ -56,9 +57,12 @@ Why Were Platforms Introduced?
 What Are Install Targets?
 ^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Install targets represent file systems. More than one platform can use the
-same file system. Cylc relies on the site configuration file ``global.cylc`` to determine
-which platforms share install targets.
+An install target is an arbitrary label that represents a file system. All
+platforms that see the same filesystem should specify the same install target
+so that Cylc knows where it needs to install workflow files.
+
+If you log into one system and see the same files as on another, then these two
+platforms will require the same install target in ``global.cylc`` config file.
 
 Cylc will setup each remote install target once. During setup it will:
 
@@ -66,12 +70,9 @@ Cylc will setup each remote install target once. During setup it will:
   - Symlink directories
   - Copy authentication keys (to allow secure communication)
 
-Note, if missing from configuration, the install target will default to the
-platform name. If incorrectly configured, this will cause errors in
-:ref:`RemoteInit`.
-
-If you log into one system and see the same files as on another, then these two
-platforms will require the same install target in ``global.cylc`` config file.
+Note, if missing from the configuration, the install target will default to
+the platform name. If incorrectly configured, this will cause errors in
+:ref:`RemoteInit` due to Cylc finding files there already during setup.
 
 Example Platform Configurations
 -------------------------------
--- a/src/reference/config/writing-platform-configs.rst
+++ b/src/reference/config/writing-platform-configs.rst
@@ -57,7 +57,7 @@ Why Were Platforms Introduced?
 What Are Install Targets?
 ^^^^^^^^^^^^^^^^^^^^^^^^^
 
-An install target is an arbitrary label that represents a file system. All
+An install target is an arbitrary label to identify a distinct file system. All
 platforms that see the same filesystem should specify the same install target
 so that Cylc knows where it needs to install workflow files.
 


src/reference/config/writing-platform-configs.rst
```diff
@@ -56,22 +57,22 @@ Why Were Platforms Introduced?
 What Are Install Targets?
 ^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Install targets represent file systems. More than one platform can use the
-same file system. Cylc relies on the site configuration file ``global.cylc`` to determine
-which platforms share install targets.
+An install target is an arbitrary label that represents a file system. All
```
MetRonnie: (Take or leave this suggestion)
```suggestion
An install target is an arbitrary label to identify a distinct file system. All
```

MetRonnie (approved):
Thanks, I was unclear on this myself

MetRonnie (changes requested):
Actually could you stick this on 8.6.x base branch

hjoliver:
@oliver-sanders or @dpmatthews - @MetRonnie has approved this, but can one of you comment on my question above re use of `install target = localhost` in the other places in the docs?

src/reference/config/writing-platform-configs.rst
```diff
@@ -41,7 +41,8 @@ Platforms define settings, most importantly:
 
  - A set of ``hosts``.
  - A ``job runner`` (formerly a ``batch system``) where Cylc can submit a job.
- - An ``install target`` for Cylc to install job files on.
+ - An ``install target``, representing each distinct filesystem that Cylc might
+   need to install workflow files on.
```
dpmatthews, line 45: Suggestion:
`An ``install target``, representing the ``$HOME`` directory used on the platform.`

src/reference/config/writing-platform-configs.rst
```diff
@@ -56,22 +57,22 @@ Why Were Platforms Introduced?
 What Are Install Targets?
 ^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Install targets represent file systems. More than one platform can use the
-same file system. Cylc relies on the site configuration file ``global.cylc`` to determine
-which platforms share install targets.
+An install target is an arbitrary label to identify a distinct file system. All
+platforms that see the same filesystem should specify the same install target
+so that Cylc knows where it needs to install workflow files.
```
dpmatthews, line 62: Suggestion:
`An install target is an arbitrary label to identify the $HOME directory. All
platforms that share the same $HOME directories should specify the same install target
so that Cylc knows whether it needs to install workflow files.`

src/reference/config/writing-platform-configs.rst
```diff
@@ -56,22 +57,22 @@ Why Were Platforms Introduced?
...
 
-Note, if missing from configuration, the install target will default to the
-platform name. If incorrectly configured, this will cause errors in
-:ref:`RemoteInit`.
-
-If you log into one system and see the same files as on another, then these two
-platforms will require the same install target in ``global.cylc`` config file.
+Note, if missing from the configuration, the install target will default to
+the platform name. If incorrectly configured, this will cause errors in
+:ref:`RemoteInit` due to Cylc finding files there already during setup.
```
dpmatthews, line 75: Suggestion:
`Note, if missing from the configuration, the install target will default to
the platform name (which is fine if there are no other platforms using the
same $HOME directories). However, if incorrectly configured (i.e. you use
platforms which share $HOME directories but use different install targets),
this will cause errors in
:ref:`RemoteInit` due to Cylc finding files there already during setup.`

dpmatthews (changes requested):
I now think it's clearer, and more accurate, if we talk about $HOME directories rather than filesystems (for example, you could have platforms which share the same network filesystem but use different $HOME directories on that filesystem).

I also think we need a note in the "Symlinking Directories" section:
`Note: If multiple platforms share the same install target and you specify a symlink directory such as ``log = $DATADIR`` then it is essential that ``$DATADIR`` is defined on all the platforms and points to the same filesystem.`

dpmatthews:
> @oliver-sanders or @dpmatthews - @MetRonnie has approved this, but can one of you comment on my question above re use of `install target = localhost` in the other places in the docs?

In the examples I've looked at, `install target = localhost` is required unless you also change the install target used for the `localhost` platform (which may add to any confusion?).

hjoliver:
> I now think it's clearer, and more accurate, if we talk about $HOME directories rather than filesystems (for example, you could have platforms which share the same network filesystem but use different $HOME directories on that filesystem).

@dpmatthews OK I hadn't thought of that. Do you know of actual cases of this - different platforms with different home paths on the same filesystem??

Anyhow, "talking about $HOME directories rather than filesystems" could also be confusing. It's common to have the same username and home dir path on different platforms (with separate filesystems) and I'm pretty sure many users would think of that as "having the same home directory" on those platforms.  Whereas, they are really "different home directories with the same path". What's the easiest way to explain this clearly? Maybe we need to talk about both filesystems and home directories to avoid any misunderstandings 🤔

hjoliver:
> In the examples I've looked at, install target = localhost is required unless you also change the install target used for the localhost platform (which may add to any confusion?).

Yes, we'd have to make the changes self-consistent of course. I just think that our frequent use of`install target = localhost` strongly gives the impression that the install target is an actual host (perhaps, the particular host on the platform that Cylc should use for installing files - which it is not).

dpmatthews:
> Do you know of actual cases of this - different platforms with different home paths on the same filesystem??

Possibly once in the past.

src/reference/config/writing-platform-configs.rst
dpmatthews: or maybe:
`An install target is an arbitrary label to identify the distinct filesystem used to store the $HOME directories. All platforms that share the same $HOME directories (i.e. the same files) should specify the same install target so that Cylc knows whether it needs to install workflow files
`

PR #932, opened 2026-05-22
GH Actions: split linkcheck into its own job

MetRonnie:
Linkcheck is the slowest part of the build and it sometimes fails due to temporary server errors etc.

**Requirements check-list**

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,10 +25,18 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+env:
+  SPHINXOPTS: '-Wn --keep-going'
+  FORCE_COLOR: 'true'
+  PIP_PROGRESS_BAR: 'off'
+
 jobs:
   test:
     runs-on: ubuntu-latest
     timeout-minutes: 10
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -66,29 +74,35 @@ jobs:
...
-          import enchant
-          print(enchant.list_dicts())
+          make html spelling
 
-      - name: build & test
+      - name: linkcheck
+        if: matrix.type == 'linkcheck'
         run: |
-          make html spelling linkcheck \
-            SPHINXOPTS='-Wn --keep-going' FORCE_COLOR=true
```
oliver-sanders, line 85: These are `make` variables, not environment variables.

Makefile
```diff
@@ -2,7 +2,7 @@
 
 # You can set these variables from the command line:
 # e.g. $ make html SPHINXOPTS=-W
-SPHINXOPTS = -n
+SPHINXOPTS ?= -n
```
MetRonnie, line 5: @oliver-sanders This makes it fall back to the `SPHINXOPTS` env var if not specified in the `make` command

oliver-sanders (approved):

PR #937, opened 2026-06-01
Improve Jinja2 docs

MetRonnie:
- Add note on what can be imported, after discovering some cases of non standard lib imports at the Met Office
- Add macros example from https://cylc.discourse.group/t/tip-jinja2-macros/1321/1
- Small changes to organisation and readability 

**Requirements check-list**


samuel-denton (approved):
LGTM

PR #944, opened 2026-06-18
Fix a typo in shared access doc.

hjoliver:
Example shared auth config does not work as currently written.

**Requirements check-list**


oliver-sanders (approved):
This is likely leftover from https://github.com/cylc/cylc-doc/pull/833/changes

PR #958, opened 2026-07-24
jinja2: improve docs for globals, filters and tests

oliver-sanders:
These docs are a little bit jumbled, these changes aim to:

* Clarify what globals/filters/tests actually are.
* Avoid the over-emphasis on filters which appears to have lead users to prefer these to globals.
* Clarify what built-ins are provided by Cylc/Jinja2.
* Provide examples of how to write custom globals/filters/tests.
* Centralise docs for Cylc builtins, make them all linkable and vaguely standardise the way they are presented.

**Requirements check-list**
- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.

PR #966, opened 2026-09-17
Clarify J2 extensions

MetRonnie:
Follow-up to #958, which was merged while review was still outstanding.

@samuel-denton If you think this is an improvement, 1 review should do, otherwise close it if it isn't an improvement

**Requirements check-list**


samuel-denton (approved):
LGTM

